PR #2, opened 2017-02-21
Add image with NeptuneUI

DunderRoffe:
Adds an image that includes NeptuneUI. In order to keep a qt free version of the Pelux reference this is done as a new image.

### DIFF
--- a/recipes-qt/automotive/neptune-ui/neptune.service
+++ b/recipes-qt/automotive/neptune-ui/neptune.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Neptune
+After=dbus.service dbus-session.service systemd-user-sessions.service
+
+[Service]
+ExecStart=/usr/bin/appman -r -c /opt/am/config.yaml -c am-config.yaml --dbus session Main.qml -platform eglfs
+Restart=on-failure
+WorkingDirectory=/opt/neptune
+Environment=QT_IM_MODULE=qtvirtualkeyboard
+Environment=XDG_RUNTIME_DIR=/tmp
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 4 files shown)


recipes-qt/automotive/neptune-ui/neptune.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Neptune
+After=dbus.service dbus-session.service systemd-user-sessions.service
+
+[Service]
+ExecStart=/usr/bin/appman -r -c /opt/am/config.yaml -c am-config.yaml --dbus none Main.qml -platform eglfs
```
GordanM: NeptuneUI should have dbus enabled by default: 

`ExecStart=/usr/bin/appman -r -c /opt/am/config.yaml --dbus session -c am-config.yaml Main.qml`

recipes-qt/automotive/neptune-ui/neptune.service
```diff
@@ -0,0 +1,13 @@
...
+Restart=on-failure
+WorkingDirectory=/opt/neptune
+Environment=QT_IM_MODULE=qtvirtualkeyboard
```
GordanM: This is only true if the qtvirtualkeyboard is a part of PELUX.

recipes-qt/automotive/neptune-ui/neptune.service
GordanM: Also, I'll get back to you on another curiosity I have regarding the service startup.

recipes-qt/automotive/neptune-ui/neptune.service
DunderRoffe: qtvirtualkeyboard is an RDEPENDS on the neptuneui recipe in boot2qt

DunderRoffe:
(assuming you are speeking of the .bbappend) Yes I agree.

PR #7, opened 2017-05-03
neptune-ui: update neptune.service to work with updated meta-boot2qt

erikboto:
Updates the systemd service for neptune so that it works on Pelux after it has been bumped to morty, https://github.com/Pelagicore/pelux-manifests/pull/35.

- Neptunes am-config.yaml is not depending on AppMans config.yaml anymore.
- Set a UTF-8 locale since AppMan will otherwise fail to find one since it doesn't have UTF-8 in its name.


DunderRoffe:
Will this merge once https://github.com/Pelagicore/pelux-manifests/pull/35 is merged

PR #9, opened 2017-05-15
core-image-pelux: add GENIVI components

GordanM:
The following GENIVI components are added:
 * dlt-daemon
 * node-state-manager
 * persistence-client-library

### DIFF
--- a/recipes-core/images/core-image-pelux.bb
+++ b/recipes-core/images/core-image-pelux.bb
@@ -12,4 +12,4 @@ IMAGE_INSTALL += "softwarecontainer"
 # GENIVI components
 IMAGE_INSTALL += " dlt-daemon         \
                    dlt-daemon-systemd \
-                   "
+                 "
--- a/recipes-core/images/core-image-pelux.bb
+++ b/recipes-core/images/core-image-pelux.bb
@@ -1,9 +1,15 @@
 #
-#   Copyright (C) 2016 Pelagicore AB
+#   Copyright (C) 2017 Pelagicore AB
 #
+
 DESCRIPTION = "Image for creating a Pelux image"
 
 require recipes-core/images/core-image-bistro-dev.bb
 
 # Pelux components
 IMAGE_INSTALL += "softwarecontainer"
+
+# GENIVI components
+IMAGE_INSTALL += " dlt-daemon         \
+                   dlt-daemon-systemd \
+                   "


GordanM:
Open question: Would we rather include the whole bulk of GENIVI components we specify in meta-bistro, as I've suggested now, or would it be better to specify component by component inside the image recipe? Please discuss. 

Also, this commit depends on a patch to meta-ivi (pending) because the dlt-daemon has dependencies to gzip.bbclass which was falsely removed.

rpannek:
In my opinion we should do it component by component, the PCL for example is something we don't really want I think, because of it's bad shape.

DunderRoffe:
Since these are components that will most likely still be required even after bootstrapping the project they should not be defined in meta-pelux. Is it still the case that meta-bistro, or at least a version of it, is what will be left in any production project?

I agree with @rpannek in principal. However, we should be careful how much customization we allow/offer since we need to test it. Also PCL might be P1 specific, meaning we have to have it. But in that case we should also already have it from meta-ivi. @jeremiah could you confirm if this is the case?

rpannek:
It's one thing to have it available in a layer but a different thing to install and configure it on the image.

jeremiah:
Also, we'll want to determine if we want to be GENIVI compliant. If we want to be GENIVI compliant, we *have* to have things like PCL, but my understanding is that we want to be more flexible and strict GENIVI compliance is not a requirement. In this case I think that each component should be brought into the system based on need and quality. If the component fits our need and its in decent enough shape to meet our need then I think we should adopt it, if not, then perhaps adopt another. Basically we treat GENIVI like any other upstream.

gillesfernandez:
@jeremiah that's the spirit :)

GordanM:
Thanks for the constructive comments everyone! 
Now a bonus question, something that @DunderRoffe mentioned. The GENIVI components we want to pull in should be defined in meta-bistro in my opinion. That is where we make the decision what type of GENIVI components our PELUX distro will include.

Does that sound right to the rest of you?

rpannek:
Is there a image recipe in meta-bistro which we inherit into PLUX? I don't feel that we should inflate the meta-bistro image for the sole purpose of inheriting it in the PLUX baseline.

GordanM:
@rpannek the current case is that these image definitions already exists in meta-bistro, with extra components that PELUX might not need like PCL (https://github.com/Pelagicore/meta-bistro/blob/master/recipes-core/images/core-image-bistro-genivi.inc). 

On the other hand, why would we want to maintain these images in meta-bistro when we decided to use meta-pelux as the core layer? Because of that, I propose either to: remove the image recipes from meta-bistro or keep them as aggregates for development and/or GENIVI recipes as we have done so far (meaning removing PCL from the meta-bistro image recipe, until we decide otherwise).

DunderRoffe:
Actually I changed my mind. We should move this to meta-pelux and remove it from meta-bistro. Since the image in meta-bistro wont be the one used anyway there is no point in keeping it there. However, PELUX still needs to be GENIVI compliant so we should provide them in core-image-pelux directly or something along those lines. This includes PCL until we have found a better version of it.

TL;DR
Keep the content but move it to meta-pelux

GordanM:
So the decision has been made: I'll be sure to remove the images from the meta-bistro layer and move the components currently found in the aggregate to this core image recipe.

PR #11, opened 2017-05-18
core-image-pelux: add node-state-manager

GordanM:


DunderRoffe:
Do you know if application manager setup connect to NSM?

GordanM:
Currently there is no integration between any of our software components and the NSM.

PR #12, opened 2017-05-22
base-files: use default fstab

GordanM:
By default mount /dev/root as a read/write partition.

Singed-off-by: Gordan Markuš <gordan.markus@pelagicore.com>


GordanM:
@DunderRoffe the idea is to have this default fstab setup in the meta-pelux layer so we avoid copying it around in the `meta-pelux-bsp-*` layers. 

In case some hardware needs a special fstab they would actually introduce it in its own BSP layer. 
Agreed, this default fstab should be removed from other BSP layers.

PR #15, opened 2017-06-09
poky: Patch automated testing script and bbclasses

GordanM:
This patch should be removed once it is mainlined.

### DIFF
--- a/layers/poky/patch/0001-Fix-automated-runtime-testing-using-SystemdbootTarge.patch
+++ b/layers/poky/patch/0001-Fix-automated-runtime-testing-using-SystemdbootTarge.patch
@@ -1,0 +1,69 @@
+From 7f2988cc0e50062277908b485c3a623d025e4cf9 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Erik=20Bot=C3=B6?= <erik.boto@pelagicore.com>
+Date: Thu, 18 May 2017 15:12:56 +0200
+Subject: [PATCH] Fix automated runtime testing using SystemdbootTarget
+
+---
+ meta/classes/testimage.bbclass           | 2 +-
+ meta/lib/oeqa/controllers/masterimage.py | 4 ++--
+ meta/lib/oeqa/utils/sshcontrol.py        | 9 +++------
+ 3 files changed, 6 insertions(+), 9 deletions(-)
+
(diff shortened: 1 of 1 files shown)


rpannek:
Perhaps we should add some info on what it is for and when to use it?

layers/poky/patch/0001-Fix-automated-runtime-testing-using-SystemdbootTarge.patch
```diff
@@ -0,0 +1,69 @@
...
+diff --git a/meta/lib/oeqa/controllers/masterimage.py b/meta/lib/oeqa/controllers/masterimage.py
+index 9ce3bf803d..4981d3d425 100644
+--- a/meta/lib/oeqa/controllers/masterimage.py
++++ b/meta/lib/oeqa/controllers/masterimage.py
+@@ -108,7 +108,7 @@ class MasterImageHardwareTarget(oeqa.targetcontrol.BaseTarget, metaclass=ABCMeta
+             time.sleep(10)
+             self.power_ctl("cycle")
+         else:
+-            status, output = conn.run("reboot")
++            status, output = conn.run("{ sleep 1; reboot; } > /dev/null &")
```
zeenix: There might be an issue with this command since it's here I get the error from. I'm checking this ..

GordanM:
@rpannek I believe that is what @breakreturn will add in the SWF. Is that correct?

rpannek:
@GordanM you might be right, I guess I miss some info about it in the commit message, what is that patch for, a link to upstream where it is discussed and so on so when I find it here half a year in the future I have a fighting chance to understand why it is here and where I can follow up in tearms of removing it, etc.

breakreturn:
@rpannek @GordanM Yeah as we talked about, we need to carry the patch somewhere and the SWF how-to on automated deployment (https://github.com/Pelagicore/software-factory/pull/12) will reference this location of the patch.

GordanM:
@rpannek and @breakreturn added a link to the bugzilla which contains further links to the patchset on the OE mailing list.

layers/poky/patch/0001-Fix-automated-runtime-testing-using-SystemdbootTarge.patch
zeenix: Actually no, this change doesn't seem to break it but it doesn't help anything either.

PR #19, opened 2017-06-27
Topic/template

GordanM:


recipes-example/template-library/template-library.bb
```diff
@@ -0,0 +1,25 @@
...
+
+DESCRIPTION = "A template C/C++ library"
+HOMEPAGE = "https://github.com/Pelagicore/template-library"
+LICENSE = "MPL-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=815ca599c9df247a0c7f619bab123dad"
+
+PR = "r0"
+PV = "1.0+git${SRCREV}"
+
+SRC_URI = "gitsm://github.com/Pelagicore/template-library.git;protocol=https;branch=master"
```
rpannek: I think you made a mistake using gitsm here because the repo doesn't seem to have any submodules.

recipes-example/template-library/template-library.bb
```diff
@@ -0,0 +1,25 @@
...
+
+SRCREV = "29d6dbd859d725ee6d10686ee5449069412de2c6"
+
+inherit cmake pkgconfig
+
+S = "${WORKDIR}/git/"
+
+PACKAGES = "${PN} ${PN}-dbg ${PN}-dev"
```
rpannek: I think it would be nicer to have the variables in their own lines.

recipes-example/template-library/template-library.bb
GordanM: Oops missed this.

recipes-example/template-library/template-library.bb
GordanM: Agreed.

GordanM:
@rpannek fixed.

PR #26, opened 2017-07-14
Add SPDX-License-Identifier to recipes

GordanM:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Pelagicore
+Copyright (c) 2016-2017 Pelagicore
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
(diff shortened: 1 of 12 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Pelagicore
+Copyright (c) 2017 Pelagicore
```
rpannek: I think it should be (c) 2016-207.

conf/layer.conf
```diff
@@ -1,6 +1,8 @@
 #
-#   Copyright (C) 2016 Pelagicore AB
+#   Copyright (C) 2017 Pelagicore AB
```
rpannek: I think it should be (c) 2016-207.

GordanM: Valid points!

GordanM:
Updated.

LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Pelagicore
+Copyright (c) 2016-2017 Pelagicore
```
zeenix: I'd put this (and the other same change) to a separate patch but it's not a biggie since it's kinda related.

GordanM:
@zeenix could you suggest what to put in the log to make it better?

GordanM:
Updated.

zeenix:
Ah nevermind, it obviously is that. :)

LICENSE
jeremiah: Copyright is for the life of the author, plus 70 years. So it is not really necessary to have 2016 - 2017. The original year of copyright is sufficient.

zeenix: @jeremiah oh, good point. Never thought of it that way. Would have saved me all those updates. :D

PR #27, opened 2017-07-24
recipes-example: Move template recipes

GordanM:
The template recipes have been moved to a layer called meta-template.
The meta-template layer is used as an example layer on how to create a
new Yocto layer and integrate it in your project. The layer is hosted
on github (https://github.com/Pelagicore/meta-template).


sashko:
Technically, you remove them with this commit.

GordanM:
@sashko I've been contemplating the same thing. But then while writing the description I've realized that I will be using the verb "move", since the recipes can be found, in the exact same state, in another layer.

PR #28, opened 2017-08-08
README.md: Add layer rationale and content explanation

GordanM:


GordanM:
@DunderRoffe we do but for consistency the lines were not broken in this change set as it was previously not done in the same file.

Furthermore, appendment is a word and I've over that already with @zeenix :) (https://en.wiktionary.org/wiki/appendment). An append file makes it an appendment.

PR #29, opened 2017-08-16
distro: Introduce "pelux" as distro name

jonte:
This makes it easier to identify the distribution running on a target
system, and it also makes it easier to identify compatible packages in
the build directory of a system.

This patch is meant to go into the Pyro release of PELUX (Yocto 2.3).

Signed-off-by: Jonatan Pålsson <jonatan.palsson@pelagicore.com>

### DIFF
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -6,6 +6,9 @@
 require conf/distro/poky.conf
 
 MAINTAINER = "Gordan Markuš <gordan.markus@pelagicore.com>"
+DISTRO = "pelux"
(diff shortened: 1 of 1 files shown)


conf/distro/pelux.conf
```diff
@@ -6,6 +6,9 @@
 require conf/distro/poky.conf
 
 MAINTAINER = "Gordan Markuš <gordan.markus@pelagicore.com>"
+DISTRO = "pelux"
+DISTRO_NAME = "PELUX (Built on on Yocto 2.3)"
```
rpannek: Is it possible to get the Yocto version number from a variable so we don't need to think of updating it every half a year?

Issue #34, opened 2017-08-17
Make core-image-pelux into a bbclass


jonte:
This will allow us to "inherit core-image-pelux" in PELUX-based images.

See for example how "core-image" works in relation to, for example, "core-image-minimal":

core-image: https://github.com/openembedded/openembedded-core/blob/master/meta/classes/core-image.bbclass
core-image-minimal: http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/images/core-image-minimal.bb?h=master

PR #35, opened 2017-08-29
core-image-pelux: turn recipe into a class

mardy:
Modify meta-pelux to make the "core-image-pelux" recipe into a bbclass,
rather than a recipe.
Having core-image-pelux be a bbclass rather than a recipe will allow us
to use the "inherit" directive in our image files, rather than the
"require" directive. "inherit" is preferred, since it does not require
us to specify the path to the bbclass.

Update the existing recipes to work with this change.

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -3,8 +3,6 @@
 #   SPDX-License-Identifier: MIT
 #
 
-DESCRIPTION = "Image for creating a Pelux image"
-
 inherit core-image
 
 # Pelux components
--- a/recipes-core/images/core-image-pelux-minimal.bb
+++ b/recipes-core/images/core-image-pelux-minimal.bb
@@ -1,0 +1,8 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+DESCRIPTION = "Image for creating a Pelux image"
+
+inherit core-image-pelux
--- a/recipes-core/images/core-image-pelux-qtauto.bb
+++ b/recipes-core/images/core-image-pelux-qtauto.bb
@@ -5,7 +5,7 @@
 
 DESCRIPTION = "Image for creating a Pelux image with Qt frontend"
 
-require recipes-core/images/core-image-pelux.bb
+inherit core-image-pelux
 
 # Pelux components
 IMAGE_INSTALL += " neptune-ui        \


GordanM:
@zeenix care to comment on this commit message? :)

GordanM:
Fixes #34 .

classes/core-image-pelux.bbclass
```diff
@@ -40,3 +38,4 @@ toolchain_create_sdk_env_script_append() {
 IMAGE_ROOTFS_SIZE ?= "1000000"
 
 IMAGE_FSTYPES ?= "ext3 sdcard"
+
```
zeenix: Unrelated change, please remove (from this patch).

classes/core-image-pelux.bbclass
```diff
@@ -3,8 +3,6 @@
 #   SPDX-License-Identifier: MIT
 #
 
-DESCRIPTION = "Image for creating a Pelux image"
-
```
zeenix: Unrelated change, please remove (from this patch).

zeenix:
@GordanM I guess you meant that Alberto should add `Fixes #34` to the commit log?

classes/core-image-pelux.bbclass
mardy: This one is not unrelated. The diff is misleading because git detects the addition of this new file as a move from the core-image-pelux.bb, since the contents are indeed mostly the same.

But in fact, I didn't remove these lines: simply didn't copy them into the new file, as they should not be added there.

zeenix: ah, the move wasn't obvious to me.

GordanM:
@zeenix the `Fixes #34` is not needed in the commit log. I've just linked the issues by stating #34 here in the comments.

zeenix:
didn't we decide on having some sort of trace in the commit logs to issues?

mardy:
I update che branch. I added "Fixes #34" in the commit message too, I also think it's a good to have it there.

GordanM:
@breakreturn can you please check the comments above regarding the issue tracing?

breakreturn:
@GordanM Most discussions that have been difficult to resolve on this topic have been when the situation involves an internal issue tracker and referencing that in the commit. That's not something I'd say we should do.

In this case, where we use the open issue tracker I don't particularly mind referencing it. However, I do not think we have any solid strategy or plan wrt exactly how to use the open issue tracker or reference it so I can't say what the consensus is here. This is fine for me as it is.

GordanM:
Thanks @breakreturn, essentially that is the same thing I've told to @zeenix IRL: The majority of our backlog is found in an internal issue tracker and that is the reason why I find having references to github issues slightly redundant, as there is no common pattern.

PR #36, opened 2017-08-31
distro: introduce "process-containment" feature

### LINKED ISSUE #33: Make softwarecontainer optional by using a distro feature
By making a distro feature for softwarecontainer, and thus making it optional for inclusion, we no longer have to rely on meta-virtualization. meta-virtualization depends on meta-python, meta-filesystems and meta-networking - so we would remove 4 required layers for PELUX users not needing softwarecontainer if we make it optional.

Current softwarecontainer inclusion is here:

https://github.com/Pelagicore/meta-pelux/blob/3b80b2bb6effe524577dde63c26081bb75a95df8/recipes-core/images/core-image-pelux.bb#L11

mardy:
Make the softwarecontainer package optional, by adding a DISTRO_FEATURE
flag called "process-containment".  In order to keep compatibility with
the existing images, this feature is enabled by default for the "pelux"
images.

Distributions which disable this feature will not have the
softwarecontainer package installed, and therefore might not require the
meta-virtualization layer, which is being pulled in just to satisfy the
softwarecontainer dependencies.

Fixes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Dependencies
 This layer depends on:
 * poky (http://git.yoctoproject.org/cgit/cgit.cgi/poky)
 * meta-openembedded (http://cgit.openembedded.org/meta-openembedded/)
-* meta-virtualization (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization)
+* meta-virtualization (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization), if the "process-containment" distro feature has been enabled
 * meta-bistro (https://github.com/Pelagicore/meta-bistro)
 
 Branching
--- a/recipes-qt/automotive/qtapplicationmanager-noop.inc
+++ b/recipes-qt/automotive/qtapplicationmanager-noop.inc
@@ -1,0 +1,6 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+# this file is intentionally left blank
--- a/recipes-qt/automotive/qtapplicationmanager-noop.inc
+++ b/recipes-qt/automotive/qtapplicationmanager-noop.inc
@@ -3,4 +3,9 @@
 #   SPDX-License-Identifier: MIT
 #
 
-# this file is intentionally left blank
+# this file is intentionally left blank: it is sourced by the _append recipe for
+# qtapplicationmanager when the softwarecontainer plugin is disabled. This file
+# is needed because the "require" directive would fail otherwise; and on the
+# other hand, we want to use the "require" directory (instead of just
+# "include") because we want an error to be generated if -- for any reason --
+# the qtapplicationmanager-sc.inc file is not found.
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -6,8 +6,12 @@
 inherit core-image
 
 # Pelux components
-IMAGE_INSTALL += "softwarecontainer"
 IMAGE_INSTALL += "packagegroup-bistro-utils"
+
+# Include softwarecontainer only if the process-containment feature has been enabled
+IMAGE_INSTALL += "\
+    ${@bb.utils.contains("DISTRO_FEATURES", "process-containment", "softwarecontainer", "", d)} \
+"
 
 # Pelux templates
 IMAGE_INSTALL += "template-service"
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -21,6 +21,7 @@ DISTRO_FEATURES_append = " \
                           largefile    \
                           opengl       \
                           pcmcia       \
+                          process-containment \
                           usbgadget    \
                           usbhost      \
                           wifi         \
--- a/recipes-qt/automotive/qtapplicationmanager-sc.inc
+++ b/recipes-qt/automotive/qtapplicationmanager-sc.inc
@@ -1,0 +1,27 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+RDEPENDS_${PN} += "${PN}-softwarecontainer"
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+EXTRA_QMAKEVARS_PRE += "\
+    -config enable-examples \
+"
+
+SRC_URI += " \
+    file://sc-config.yaml \
+    "
+
+do_install_append_class-target() {
+    install -d ${D}${libdir}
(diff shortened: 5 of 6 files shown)


recipes-qt/automotive/qtapplicationmanager-noop.inc
```diff
@@ -0,0 +1,6 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+# this file is intentionally left blank
```
rpannek: Could you add a comment about the purpose why you added this file?

jonte: I guess it is to make the require line work in qtapplicationmanager_git.bbappend. I think it should be possible to do this in some other way, but I am not sure..

jonte:
This mostly looks good to me. I'm not too fond of the noop hack though.

Also, a more general point: The purpose of doing a DISTRO_FEATURE is mostly that we want to be able to enable the softwarecontainer recipe *and* some other recipes which it depends on - such as patches to the kernel for cgroups, namespaces, virtual ethernet support, etc.

I am not sure how we should bundle these "features" together to make the DISTRO_FEATURE complete.. Maybe we need to collect all these recipes & bbappends in some packagegroup..?

This general concern doesn't need to be solved in this first step, but I think it is important to keep in mind.

I am OK with merging this after we have either fixed the noop thing, or concluded that this is the only way to do a "conditional require".

recipes-qt/automotive/qtapplicationmanager-noop.inc
mardy: That's correct. But I will add a comment within the file itself to explain what's going on.

GordanM:
I believe these two commits should have been squashed instead of producing two separate commits. @zeenix can you please comment on this? :)

README.md
```diff
@@ -12,7 +12,7 @@ Dependencies
 This layer depends on:
 * poky (http://git.yoctoproject.org/cgit/cgit.cgi/poky)
 * meta-openembedded (http://cgit.openembedded.org/meta-openembedded/)
-* meta-virtualization (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization)
+* meta-virtualization (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization), if the "process-containment" distro feature has been enabled
```
zeenix: this line is getting way too long. I generally prefer lines to be <= 80 chars but we don't have to follow such a short limit (120 maybe?).

jonte:
Uhh.. I thought I did squash.. I used the GitHub UI, maybe I misread the button. I can force-push a squashed version if you're all OK w/ it.

recipes-qt/automotive/qtapplicationmanager-noop.inc
```diff
@@ -3,4 +3,9 @@
 #   SPDX-License-Identifier: MIT
 #
 
-# this file is intentionally left blank
+# this file is intentionally left blank: it is sourced by the _append recipe for
+# qtapplicationmanager when the softwarecontainer plugin is disabled. This file
+# is needed because the "require" directive would fail otherwise; and on the
+# other hand, we want to use the "require" directory (instead of just
+# "include") because we want an error to be generated if -- for any reason --
+# the qtapplicationmanager-sc.inc file is not found.
```
zeenix: Fixes to unmerged patches, should always been squashed into the original patches. i-e this patch should be squashed into previous patch.

recipes-qt/automotive/qtapplicationmanager-noop.inc
```diff
@@ -3,4 +3,9 @@
 #   SPDX-License-Identifier: MIT
 #
 
-# this file is intentionally left blank
+# this file is intentionally left blank: it is sourced by the _append recipe for
```
zeenix: Use of ":" is wrong here AFAICT. LIke @jonte, i don't like this hack either FWIW.

recipes-qt/automotive/qtapplicationmanager-noop.inc
```diff
@@ -3,4 +3,9 @@
...
+# qtapplicationmanager when the softwarecontainer plugin is disabled. This file
+# is needed because the "require" directive would fail otherwise; and on the
+# other hand, we want to use the "require" directory (instead of just
+# "include") because we want an error to be generated if -- for any reason --
```
zeenix: "-- for any reason --" -> ", for any reason, " or put it in brackets.

README.md
mardy: That would be my preference as well, but this file obviously doesn't follow that rule, given that all paragraphs are on a single line. I'll change this anyway.

recipes-qt/automotive/qtapplicationmanager-noop.inc
mardy: Please elaborate on the ":". I think it's correct to use a colon to clarify the previous statement, but I'm open to learn something new :-)

As for the hack, I fully agree with you, but I couldn't come up with a better alternative. I've googled for a day and this was the best I could find :-)

zeenix: It would be correct if the phrase before ':' would suggest that something is to follow. E.g "The reason for leaving this file blank is: ".

Oh and please please always capitalize the first letter in text that's supposed to be any European language. :)

zeenix: About the hack, I guess if @jonte wasn't able to suggest a better solution, there is not likely a better way. :)

README.md
mardy: Created https://github.com/Pelagicore/meta-pelux/pull/37 for this.

mardy:
Created https://github.com/Pelagicore/meta-pelux/pull/38; given that this PR has already been merged, please add any additional comments into that one.

zeenix:
bah, i didn't realize it was merged prematurely. bad @jonte

rpannek:
That's why I mostly suggest the original author should merge because they know the timing of the merge best.

recipes-qt/automotive/qtapplicationmanager-noop.inc
mardy: I've fixe the capitalization on the other PR. As for the colon, I believe it's not wrong, given how it's used in this page: https://www.grammarbook.com/punctuation/colons.asp (search for "judgement call" in the text)

zeenix: that's for the rule "A colon instead of a semicolon may be used between independent clauses when the second sentence explains, illustrates, paraphrases, or expands on the first sentence." where as in here it's more like "phrase: reason" instead of "phrase : explanation/expansion of phrase". Anyway, getting really nitpicky now. :D No biggie anyway.

PR #39, opened 2017-09-06
Fix appman deps on aarch64

jonte:
These patches are needed to run Neptune on Qt ApplicationManager on Qt 5.9.2 on aarch64

### DIFF
--- a/recipes-temporary-patches/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
+++ b/recipes-temporary-patches/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
@@ -2,3 +2,7 @@
 # and uses qemu for this. For aarch64, these commands cause qemu to crash, so
 # we disable introspection.
 EXTRA_OECONF_aarch64 += "--disable-introspection"
+
+# Wayland support is broken, so remove this until it is fixed upstream. Fixes:
+# | make[3]: *** No rule to make target 'viewporter-protocol.c', needed by 'all'.  Stop.
+PACKAGECONFIG_remove_aarch64 = "wayland"
--- a/recipes-temporary-patches/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
+++ b/recipes-temporary-patches/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
@@ -1,0 +1,4 @@
+# GObject introspection needs to run some commands on the native architecture,
+# and uses qemu for this. For aarch64, these commands cause qemu to crash, so
+# we disable introspection.
+EXTRA_OECONF_aarch64 += "--disable-introspection"
--- a/recipes-temporary-patches/librsvg/librsvg_%.bbappend
+++ b/recipes-temporary-patches/librsvg/librsvg_%.bbappend
@@ -1,0 +1,4 @@
+# GObject introspection for librsvg needs to run some commands on the native
+# architecture, and uses qemu for this. For aarch64, these commands cause qemu
+# to crash, so we disable introspection.
+EXTRA_OECONF_aarch64 += "--disable-introspection"


recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
```diff
@@ -0,0 +1,8 @@
+# GObject introspection needs to run some commands on the native architecture,
+# and uses qemu for this. For aarch64, these commands cause qemu to crash, so
+# we disable introspection.
+EXTRA_OECONF_aarch64 += "--disable-introspection"
+
+# Wayland support is broken, so remove this until it is fixed upstream. Fixes:
```
breakreturn: Is this part something that should go into like "temporary patches" if we believe it will be fixed? I'm not sure we still use the pattern of highlighting things as temporary though, @tobsan might know.

jonte: It's likely to be fixed at some point upstream.. I have seen discussions around it on the OE mailing list, so people are working on it.

jonte: Same goes for the introspection fixes, really.

tobsan: I'm not sure what the current process for handling these things are. The upside of using `recipes-temporary-patches` or similar is that we always know where these are at. On the other hand, they end up away from their more logical location.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
```diff
@@ -0,0 +1,8 @@
...
+# | make[3]: *** No rule to make target 'viewporter-protocol.c', needed by 'all'.  Stop.
+PACKAGECONFIG_remove = "wayland"
```
GordanM: Is this `PACKAGECONFIG_remove` valid for all architectures or just for aarch64?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
GordanM: Previously, in `meta-bistro`, we would stage these bbappends in the same locations as in their original layer instead of using a temporary patches folder. The contributor and maintainers were responsible to follow up with upstream and ensure these patches get included and the temporary patches are removed. Though this caused a lot of cruft over the years.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
jonte: I'm actually not sure.. I have only seen the error on aarch64, but OTOH, I have only built on aarch64.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
breakreturn: AFAIK we never changed the principle of "highlighting", or gathering, patches and tweaks that we see as temporary because of pending changes upstream (for example) in a common place named accordingly. The main reason was to avoid cruft. Maybe @GordanM knows about some rationale behind moving away from that, unless it was just a lack of information?

The "temporayness" of a patch was never defined, it's mostly an evaluation. I don't think that every fix should be considered temporary. If there's some explicit intent upstream to fix something, or if we have a patch that is pending merge, then the patch in the layer is temporary IMO.

I think you guys can decide what's best, I only hope we are consistent whatever the decision is. And that we document the pattern and the rationale to avoid de-facto ways of working that differs from the agreed ways of working.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
GordanM: @jonte I am assuming you are using the `master` PELUX branch (which is actually `morty`). I haven't seen this issue for Intel architectures, otherwise our current CI jobs would be failing, right @tobsan ?

Gagi2k:
I can't see how this is related to appman at all, it might be related to neptune-ui as this needs qtmultimedia, which needs gstreamer, but i can't see any other connection ?

jonte:
@Gagi2k You're right. This is for Neptune, not appman - sorry for the confused merge message. This message is not reflected in the actual commits, so that's something.. :)

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
jonte: @GordanM I'm using master of PELUX, yeah.

jonte: I should note that I am using gstreamer1.0 from poky, and not from meta-ivi - maybe that's a difference? I'm using gstreamer1.0-plugins-bad version 1.10.4-r0

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
tobsan: Yeah, i would actually vote for using the temporary patches directory way of doing this, just so that one can easily find everything that is supposed to be synced/patched/fixed upstream eventually.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
tobsan: Yes, If this was an issue we would already have failures. For that reason, I would like this to be changed so that it only applies to aarch64 unless we KNOW it is needed for all archs.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
jonte: Alright, I will send a v2 and move this bbappend into a new recipe directory called "recipes-temporary-patches/gstreamer/" - sounds OK?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
jonte: Alright, I'll quality the `PACKAGECONFIG_remove` w/ `_aarch64`

PR #40, opened 2017-09-06
README.md: describe relation with meta-bistro

tobsan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The layer provides the PELUX distro that is derived from poky. Furthermore,
 this layer provides `core-image-pelux` recipes, along with necessary additions
 to software components to enable a functional In-Vehicle Infotainment (IVI)
 base image. The software component additions and appendments found in this
-layer are specific for the PELUX image but can be used outside of this context. 
+layer are specific for the PELUX image but can be used outside of this context.
 
 Maintained at https://github.com/pelagicore/meta-pelux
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,9 +18,14 @@ This layer depends on:
 * poky (http://git.yoctoproject.org/cgit/cgit.cgi/poky)
 * meta-openembedded (http://cgit.openembedded.org/meta-openembedded/)
 * meta-virtualization (http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization),
-if the "process-containment" distro feature has been enabled
+  if the "process-containment" distro feature has been enabled
 * meta-bistro (https://github.com/Pelagicore/meta-bistro)
 
+The relationship with meta-bistro, which is also developed at Pelagicore, is
```
GordanM: Should the layer name be in back ticks (`meta-bistro`)? Have to check how its done in the other parts of the SWF.

README.md
```diff
@@ -18,9 +18,14 @@ This layer depends on:
...
+that meta-bistro contains general software recipes, but no distribution and no
```
GordanM: I would add that `meta-bistro` adds IVi and development related recipes and changes to the PELUX image. For instance it adds the development tools packageconfig and the ivi packageconfig (have to verify this).

tobsan: meta-bistro does not change the PELUX image. Rather, the PELUX image makes use of things defined in meta-bistro.

I just checked, and it does create/append to some packagegroups, but the relation is still that PELUX  uses what's defined in bistro.

tobsan: That said, I think the comment about describing what it is that bistro brings rather than generic software, is a good one :)

tobsan:
@GordanM updated after your comments.

PR #42, opened 2017-09-11
qtapplicationmanager: Fix expansion in inc file

jonte:
The ${PN} variable in qtapplicationmanager-sc.inc does not appear to
expand properly when it is being assigned using the := operator, so
expand it manually

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/qtapplicationmanager-sc.inc
+++ b/layers/b2qt/recipes-qt/automotive/qtapplicationmanager-sc.inc
@@ -4,7 +4,10 @@
 #
 
 RDEPENDS_${PN} += "${PN}-softwarecontainer"
-FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+# Due to immediate expansion of variables using :=, ${PN} is not yet available
+# in this .inc-file
+FILESEXTRAPATHS_prepend := "${THISDIR}/qtapplicationmanager:"
 
 EXTRA_QMAKEVARS_PRE += "\
     -config enable-examples \


layers/b2qt/recipes-qt/automotive/qtapplicationmanager-sc.inc
```diff
@@ -4,7 +4,10 @@
 #
 
 RDEPENDS_${PN} += "${PN}-softwarecontainer"
-FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+# Due to immediate expansion of variables ${PN} is not yet available in this
+# .inc-file
```
mardy: This comment is a bit misleading, the ${PN} variable is used just a couple of lines above. Better rephrase the comment, to mention that it's about the ":=" assignment. (but then, can't we just use "=" instead?)

jonte: Good point! I will rephrase. No, we need to use := for `${THISDIR}` to work, I believe. I'll check though!

jonte: Yeah, = doesn't expand properly. We need to use `:=`.

jonte: It is pretty conventional to use either ${PN}, or "files" as the name of the directory containing extra files for the recipe. I will rename the directory to "files", and thus avoid having to hardcode "qtapplicationmanager" into the inc file.

PR #43, opened 2017-09-11
qtivi: Fix building without ivigenerator

jonte:
Temporary fix to allow building qtivi without the ivigenerator


caybro:
LGTM, if it makes the build pass :+1:

GordanM:
@jonte I approve the patch, though shouldn't the QtIVI patch be included in mainline? Seems like a generic improvement.

jonte:
@GordanM Yep, this is just a temporary fix, to be kept until mainline merges it.

GordanM:
Cool!

PR #45, opened 2017-09-15
Moves qtauto from recipe to class

tobsan:
This will make it easier to inherit for whoever wants to extend it.


classes/core-image-pelux-qtauto.bbclass
```diff
@@ -0,0 +1,21 @@
...
+DESCRIPTION = "Class for inheriting a Pelux image with Qt frontend"
+
+inherit core-image-pelux
+
+# A bunch of QtAuto components
+QTAUTO_COMPONENTS := " \
+    qtapplicationmanager \
+    qtivi \
+    qtvirtualkeyboard \
+    dbus-session \
```
GordanM: Maybe this should not be stated here explicitly but pulled in as a part of `qtappman.bbappend` dependency.

layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
```diff
@@ -0,0 +1,11 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+DESCRIPTION = "Reference PELUX image with QtAuto frontend"
+
+inherit core-image-pelux-qtauto
+
+# We use neptune as the reference UI
```
GordanM: Rename "We" to a more generic name, e.g. the image.

classes/core-image-pelux-qtauto.bbclass
```diff
@@ -0,0 +1,21 @@
...
+"
+
+# Classes are special so we only do this if b2qt (which contains the components) is enabled
```
GordanM: Maybe be more explicit and say: the b2qt meta layer is used/enabled, b2qt might not be obvious what it is to others.

classes/core-image-pelux-qtauto.bbclass
```diff
@@ -0,0 +1,21 @@
...
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+DESCRIPTION = "Class for inheriting a Pelux image with Qt frontend"
+
+inherit core-image-pelux
+
+# A bunch of QtAuto components
+QTAUTO_COMPONENTS := " \
```
jonte: `:=` is not necessary I think.. `=` should work, since there are no variables to expand?

tobsan: Right, `:=` gives immediate expansion, but it is not necessary here. I was poking around with it earlier when it didn't work, but forgot to revert.

classes/core-image-pelux-qtauto.bbclass
tobsan: You mean only dbus-session, right?

GordanM: That's correct, sorry for not being explicit.

PR #47, opened 2017-09-22
Remove template service from pelux bbclass

### LINKED ISSUE #32: Remove dependency on meta-template
I don't think it should be a required layer for all PELUX-based projects, since the template-service provided might not be used by all projects.

Dependency introduced here:

https://github.com/Pelagicore/meta-pelux/blob/3b80b2bb6effe524577dde63c26081bb75a95df8/recipes-core/images/core-image-pelux.bb#L15

tobsan:
Also adds an image with the template service installed on it.

Closes #32 


tobsan:
Actually, I just realized, it might be better to move the whole template image into a conditonal structure that only is applied if meta-template is available in the bblayers. Would that be even nicer? Then one could build a complete PELUX-based system without having to pull in meta-template to begin with.

PR #48, opened 2017-10-03
Preparations for creating pyro branch

tobsan:
When syncing with upstream layers, the qtivi patch is fixed and the gstreamer recipe has issues (they are upstreamed but not in 13.0 branch)


tobsan:
@GordanM Yes, all of these patches were necessary when syncing with upstream pyro branches (the idea is to branch off master to pyro here as well). Unfortunately, meta-ivi aren't very fast at porting the patches they get. The gstreamer fixes are in master but not in 13.0 in meta-ivi.

GordanM:
@tobsan I had a hunch this patch was familiar, it was contributed by Pelagicore but not included to the right branch. Contact the maintainer please. Until then we proceed with the temporary patch.

tobsan:
@GordanM see here: https://github.com/GENIVI/meta-ivi/pull/58

PR #49, opened 2017-10-03
bsp layers: update available image names

tobsan:
qtauto does not yet work with rpi, that's why that image isn't listed in
the conf-notes yet.

### DIFF
--- a/meta-intel-extras/conf-qt/conf-notes.txt
+++ b/meta-intel-extras/conf-qt/conf-notes.txt
@@ -1,5 +1,5 @@
 Available PELUX images:
-* core-image-pelux
-* core-image-pelux-qtauto
+* core-image-pelux-minimal
+* core-image-pelux-qtauto-neptune
 
 Build them by running: bitbake <image>
--- a/meta-intel-extras/conf/conf-notes.txt
+++ b/meta-intel-extras/conf/conf-notes.txt
@@ -1,4 +1,4 @@
 Available PELUX images:
-* core-image-pelux
+* core-image-pelux-minimal
 
 Build them by running: bitbake <image>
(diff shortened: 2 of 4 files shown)


meta-intel-extras/conf-qt/conf-notes.txt
```diff
@@ -1,5 +1,5 @@
 Available PELUX images:
-* core-image-pelux
-* core-image-pelux-qtauto
+* core-image-pelux-minimal
+* core-image-pelux-qtauto-neptune
```
GordanM: Is there a reason you have added neptune to the end here?

tobsan: Yes. With the bbclass-ification of the qtauto and pelux images, the image names have changed.

DunderRoffe: Maybe the name of the bbclass should change instead? Adding neptune is redundant and confusing and the image names will be seen by more people than the names of the bbclasses.

GordanM: Thanks Tobias. 

Come to think about it: Qt Auto is a framework and Neptune is just the reference UI so it would kinda make sense to have it explicitly in the image name. You could have a Qt Auto Pelux with your own SysUI and App implementations.

tobsan:
@DunderRoffe I'm open for better suggestions.

DunderRoffe:
So maybe call the bbclass "pelux-qtauto" since the "core-image" part is not really relevant in a bbclass setting anyway

tobsan:
Well, `core-image` is a bbclass, and the idea is to inherit the bbclass for making one's own images based on QtAS, so naming has to be good. Could we call the image `core-image-pelux-neptune` perhaps? Neptune implies QtAS anyway.

PR #50, opened 2017-10-05
cmake: add override for CMAKE_STAGING_PREFIX in SDK

mardy:
By setting this variable, all projects built against the SDK using cmake
will be installed under the SDK sysroot's "/usr" directory (unless they
override the CMAKE_INSTALL_PREFIX explicitly).

This helps when building third-party libraries: getting installed in the
standard locations under the SDK sysroot means that they can be found by
tools like pkg-config with no additional effort.

See https://github.com/Pelagicore/software-factory-blueprint/pull/17 for
the discussion which led to this choice.


mardy:
Just incase you wonder: yes, I've tested this :-)
The override file is installed under `/opt/pelux/2.3.1/sysroots/x86_64-pokysdk-linux/usr/share/cmake/OEToolchainConfig.cmake.d/StagingPrefix.cmake` and does its job.

PR #52, opened 2017-10-09
bblayers.conf: remove meta-qt5 unless QtAS

### LINKED ISSUE #51: meta-qt5 should only be in bblayers if building with QtAS
For some reason, meta-qt5 is in pelux-base.xml. It should be in pelux-qt.xml. The bblayers.conf should be updated accordingly, so that the layer is only included if building the QtAS version of PELUX

tobsan:
We only need meta-qt5 in our bblayers.conf if we want to build with Qt,
meaning that we want to build the QtAS variant of PELUX. In all other
cases we don't need any Qt support.

Closes #51 


tobsan:
@DunderRoffe updated

PR #53, opened 2017-10-13
Remove package-bistro-utils from bbclass

DunderRoffe:
Utils like vim and sftpserver should not be installed in the minimal
version of pelux.

Signed-off-by: Viktor Sjölind <viktor.sjolind@pelagicore.com>

### DIFF
--- a/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune-dev.bb
+++ b/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune-dev.bb
@@ -1,0 +1,10 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+DESCRIPTION = "Reference PELUX image with QtAuto frontend"
+
+require core-image-pelux-qtauto-neptune.bb
+
+IMAGE_INSTALL += " packagegroup-bistro-utils" 
--- a/recipes-core/images/core-image-pelux-minimal-dev.bb
+++ b/recipes-core/images/core-image-pelux-minimal-dev.bb
@@ -1,0 +1,9 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+require core-image-pelux-minimal.bb
+
+# Pelux components
+IMAGE_INSTALL += " packagegroup-bistro-utils"
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -5,9 +5,6 @@
 
 inherit core-image
 
-# Pelux components
-IMAGE_INSTALL += "packagegroup-bistro-utils"
-
 # Include softwarecontainer only if the process-containment feature has been enabled
 IMAGE_INSTALL += "\
     ${@bb.utils.contains("DISTRO_FEATURES", "process-containment", "softwarecontainer", "", d)} \


rpannek:
I kind of agree with @GordanM, the only problem is that we call it minimal and it really isn't. At the same time I'd like to keep the number of different images as low as possible.

GordanM:
What the solution then be to create another image `bbclass` that extends the minimal image `bbclass` that is really "minimal" as @rpannek suggested.

DunderRoffe:
@GordanM I am only talking about removing them from the minimal case.

DunderRoffe:
It is also about what we communicate. At least I would think that if you remove something from minimal PELUX it is no longer PELUX. That is if we keep these utils in the minimal version we are implying that if you remove for instance sftpserver from the image it is no longer PELUX. Which does not seem right to me.

GordanM:
@DunderRoffe that's true. The minimal image should indeed be as minimal as possible, though by just removing the bistro utils from the minimal image you remove them everywhere. Couldn't you create another `bbclass` that extends this class with these utilities?

breakreturn:
I can only give a fairly abstract and possibly useless comment :)

I think all images we have should have specific purposes and use-cases tied to them. That should help us define what would be included in each. Also, we do own the definition of "minimal" in this case, so if we let the needs and use-cases define what we want to create we should be good.

Do we have a description of the images (purpose, use-cases, etc) in the SWF?

DunderRoffe:
Added a core-image-pelux-dev bbclass which has the utils. Should I also add one for qtauto?

mardy:
I think that a text editor is needed even in a minimal image; I'm fine with removing `vim` if there's some other (more lightweight) text editor.
As for `sftpserver`, same thing: the functionality of moving stuff to and from the device is IMHO essential; but then, if there are other ways to achive the same goal, we can drop all but one.

GordanM:
@mardy not in a production image though, so the minimal definition here is vague. I agree with @breakreturn, define the use cases and describe what is each image class for.

DunderRoffe:
So I get your point @mardy. Although poky comes with vi so there is already an editor. I also agree that a way to transfer files is nice but I do not think one should nor can force that choice on a production project.

Yes we need to define the concrete use-case for the minimal image.
My use-case for the minimal image is to make a production-ish image based on PELUX and I encounter problems while compiling utilities, none of which I would like to have on the image. So in my use-case of the minimal PELUX they force me to do workarounds instead of adding value.

tobsan:
Earlier we have had a `-dev` and non-dev image in some projects for exactly these purposes. Sure, there will be some more images to maintain but for the most part maintaining the dev image is no work at all, since it is just the regular image plus some stuff.

DunderRoffe:
Is there any action I should take to get this PR merged or are we happy with the format?

tobsan:
I would prefer having a dev image for the regular images, extending those, instead of a bbclass that would contain the same things plus dev stuff.

GordanM:
Agree with @tobsan .

layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune-dev.bb
```diff
@@ -0,0 +1,12 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+DESCRIPTION = "Reference PELUX image with QtAuto frontend"
+
+inherit core-image-pelux-qtauto
```
tobsan: Instead `require core-image-pelux-qtauto-neptune.bb`, otherwise this becomes just a duplicate + extra packages.

recipes-core/images/core-image-pelux-minimal-dev.bb
```diff
@@ -0,0 +1,9 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+require core-image-minimal
```
tobsan: Same here, `require core-image-pelux-minimal.bb`

PR #54, opened 2017-10-19
Move variants and configurations to conf directory

tobsan:
Splits out the variant configurations from the the BSP specific layers,
so that the BSP layers only contain recipes, and the variant
configurations only contain build configuration (bblayers.conf,
local.conf etc) which makes it easier to use a common base for those
configuration files.


tobsan:
Obviously, once this gets merged the SWF will have to be updated with new instructions.

conf/variant/intel-qtauto/bblayers.conf.sample
```diff
@@ -0,0 +1,13 @@
+# These are needed for bitbake to pick up the includes
+BSPDIR := "${TOPDIR}/../"
```
mardy: The last "/" is not needed. It's fine by me, though, unless you have a policy against it.

conf/variant/intel/bblayers.conf.sample
```diff
@@ -0,0 +1,12 @@
+# These are needed in order to make bitbake pick up the includes
+BSPDIR := "${TOPDIR}/../"
+BBPATH .= ":${BSPDIR}/sources/meta-pelux"
```
mardy: I see these identical definitions in other places; maybe they could also be moved to the common conf?

conf/variant/intel-qtauto/bblayers.conf.sample
tobsan: Good point, will remove it.

conf/variant/intel/bblayers.conf.sample
tobsan: Unfortunately not, since BBPATH is the variable used for searching for the common conf. Hence it is needed in the top of any file that is a top-level bblayers.conf (which then can pull in common stuff)

tobsan:
@mardy please re-review :)

PR #56, opened 2017-10-23
Prepare the rpi for Qt Auto

tobsan:
Perhaps the qtwayland thing should be upstreamed to the qt layers, it could easily be solved with a check on MACHINE and MACHINE_FEATURES or something like that.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
+++ b/layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
@@ -10,3 +10,24 @@ RDEPENDS_${PN}_append = "\
       qtwebengine        \
       qtquickcontrols2   \
       "
+
+HAS_CONTAINMENT = "${@bb.utils.contains('DISTRO_FEATURES', 'process-containment', '-c /opt/am/sc-config.yaml', '', d)}"
+
+#
+# Add software-container AM config to appman cmdline if we have containment
+# support
+#
+do_install_prepend() {
+    sed -i -e "s|\$EXTRA_ARGUMENTS|${HAS_CONTAINMENT}|" ${WORKDIR}/neptune.service
+}
+
+#
+# If we are installing on a rpi3, we need to force wayland to use the broadcom
+# driver when starting appman with neptune.
+#
+do_install_prepend_rpi() {
+    ENV_LINE="Environment=QT_WAYLAND_HARDWARE_INTEGRATION=brcm"
+    if ! grep -q "$ENV_LINE" "${WORKDIR}/neptune.service"; then
+        sed -i -e "/^\[Install\]$/i $ENV_LINE" ${WORKDIR}/neptune.service
+    fi
+}
--- a/meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
+++ b/meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
@@ -1,0 +1,8 @@
+
+#
+# Reserve 512MB RAM for the GPU if this is the QtAS variant, otherwise 128MB
+#
+# This checks if the b2qt layer is included.
+#
+
+GPU_MEM="${@bb.utils.contains("BBFILE_COLLECTIONS", "b2qt", 512, 128, d)}"
--- a/layers/b2qt/recipes-qt/automotive/neptune-ui/neptune.service
+++ b/layers/b2qt/recipes-qt/automotive/neptune-ui/neptune.service
@@ -4,7 +4,7 @@ After=dbus-session@root.service
 Requires=dbus-session@root.service
 
 [Service]
-ExecStart=/usr/bin/appman -r -c /opt/am/sc-config.yaml -c am-config.yaml --dbus session Main.qml -platform eglfs
+ExecStart=/usr/bin/appman -r -c am-config.yaml --dbus session -platform eglfs $EXTRA_ARGUMENTS
 Restart=on-failure
 WorkingDirectory=/opt/neptune
 Environment=HOME=/home/%u/
--- a/layers/b2qt/recipes-qt/qt5/qtbase_%.bbappend
+++ b/layers/b2qt/recipes-qt/qt5/qtbase_%.bbappend
@@ -1,0 +1,5 @@
+
+#
+# Remove gbm support if we're building for rpi
+#
+PACKAGECONFIG_remove_rpi = "gbm"
--- a/layers/b2qt/recipes-qt/qt5/qtwayland_%.bbappend
+++ b/layers/b2qt/recipes-qt/qt5/qtwayland_%.bbappend
@@ -1,0 +1,4 @@
+#
+# Make sure we build with support for the broadcom driver if building for rpi
+#
+PACKAGECONFIG_append_rpi = " wayland-brcm "


meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
```diff
@@ -0,0 +1,6 @@
+
+#
+# Reserve 512MB RAM for the GPU if this is the QtAS spin, otherwise 128MB
```
DunderRoffe: I think we use the term variant instead of spin. Not sure though.

meta-rpi-extras/recipes-qt/automotive/neptune-ui/neptune.service
```diff
@@ -0,0 +1,18 @@
...
+Requires=dbus-session@root.service
+
+[Service]
+ExecStart=/usr/bin/appman -r -c /opt/am/sc-config.yaml -c am-config.yaml --dbus session Main.qml -platform eglfs
+Restart=on-failure
+WorkingDirectory=/opt/neptune
+Environment=HOME=/home/%u/
+Environment=QT_WAYLAND_HARDWARE_INTEGRATION=brcm
+Environment=LC_ALL=en_US
+Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/root/dbus/user_bus_socket
```
DunderRoffe: Isn't this line redundant?

meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
tobsan: Good point, we speak of variants in the SWF and in the `conf/` directory, so we should be consistent.

meta-rpi-extras/recipes-qt/automotive/neptune-ui/neptune.service
tobsan: Ehm...no? This points out where the session bus is at, the `--dbus session` option only tells appman to use a session bus, not which one.

DunderRoffe: Sure but systemd ususally does that automagically. Altough I now see that you specify root user

meta-rpi-extras/recipes-qt/automotive/neptune-ui/neptune.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Neptune
+After=dbus-session@root.service
+Requires=dbus-session@root.service
+
+[Service]
+ExecStart=/usr/bin/appman -r -c /opt/am/sc-config.yaml -c am-config.yaml --dbus session Main.qml -platform eglfs
```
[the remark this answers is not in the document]
tobsan: Well, that would be nice but we are relying on the systemd bbclass where one points out what the service is. We could of course add a line to the existing service file as well, with sed or something. 

Having a separate file for rPi means easier to adapt to rPi specifics, but harder to follow upstream. Just changing existing file means easier to follow upstream (meta-pelux in general, that is) but more work to add specifics.

meta-rpi-extras/recipes-qt/automotive/neptune-ui/neptune.service
```diff
@@ -0,0 +1,18 @@
...
+Restart=on-failure
+WorkingDirectory=/opt/neptune
+Environment=HOME=/home/%u/
+Environment=QT_WAYLAND_HARDWARE_INTEGRATION=brcm
+Environment=LC_ALL=en_US
+Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/root/dbus/user_bus_socket
+Environment=QT_IM_MODULE=qtvirtualkeyboard
+Environment=XDG_RUNTIME_DIR=/tmp
```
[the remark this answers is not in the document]
tobsan: Can it be removed completely? Otherwise I think we should keep it for now. This is, after all, just a showcase UI, right?

meta-rpi-extras/recipes-qt/automotive/neptune-ui/neptune.service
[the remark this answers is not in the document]
tobsan: Also, the sc-config inclusion should only be done if we have a specific DISTRO_FEATURE added. I will try to rework this file with a proper recipe.

tobsan:
@Gagi2k @DunderRoffe Updated the PR in a rather major way, please re-check :)

layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
```diff
@@ -10,3 +10,40 @@ RDEPENDS_${PN}_append = "\
...
+#
+do_install_prepend() {
+    echo "Containment: ${HAS_CONTAINMENT}"
+    SC_CONFIG_LINE="-c /opt/am/sc-config.yaml"
+    SERVICEFILE="${WORKDIR}/neptune.service"
+
+
+    # We only add it if we have containment and it isn't already there
+    if [ "${HAS_CONTAINMENT}" = "TRUE" ]; then
+        if  ! grep -q "sc-config.yaml" $SERVICEFILE; then
```
tobsan: If someone knows a better way to grep for SC_CONFIG_LINE (it interprets the -c etc in a bad way), I would prefer that.

Gagi2k: How about adding $__EXTRA_ARGUMENTS__ at the end of ExecStart in the service file, which would be resolved as an empty variable by default.

In your SED line you could then replace $__EXTRA_ARGUMENTS__ with your config + $__EXTRA_ARGUMENTS__, which would also enable other recipes to add extra options ?

In addition i'm not sure whether it's worth to check whether sc-config.yaml is already part of the line, if it added twice, it doesn't matter, as appman shouldn't have a problem with that and if a user has it's custom container-config.yaml instead it won't work anyway ?

layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
```diff
@@ -10,3 +10,40 @@ RDEPENDS_${PN}_append = "\
...
+            echo "Adding SC config line"
+            sed -i -e "s|^ExecStart=\(.*\)$|ExecStart=\1 $SC_CONFIG_LINE|" $SERVICEFILE
+        fi
+    else
+	# Remove any sc configs from the config file
+        echo "Removing SC line if present"
+        sed -i -e "s|$SC_CONFIG_LINE||" $SERVICEFILE
```
Gagi2k: Why do you need to remove it ? If the DISTRO doesn't have the feature, it shouldn't be there in the first place ? And if it is, the user might know what he is doing ?

layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
tobsan: I like the idea of using $EXTRA_ARGUMENTS actually, as it can make this much easier, one can set its value already in the HAS_CONTAINER variable and then just replace it in the sed script. And if it is already replaced then the sed line won't find $EXTRA_ARGUMENTS anyway and we're good.

layers/b2qt/recipes-qt/automotive/neptune-ui_git.bbappend
tobsan: Sure, fair point.

tobsan:
@Gagi2k updated now, much cleaner and shorter :)

PR #58, opened 2017-10-25
conf/layer.conf: add a variable with the URL of the PELUX git repo

mardy:
This variable can be used by this layer's recipes like this:

SRC_URI = "${PELUX_GIT_URL}/project"

in order to continue working unmodified when the URL of the server
changes.

Signed-off-by: Alberto Mardegan <amardegan@luxoft.com>

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -6,6 +6,9 @@
 # We have a conf and classes directory, append to BBPATH
 BBPATH .= ":${LAYERDIR}"
 
+# URL of the git server hosting PELUX projects
+PELUX_GIT_URL ?= "https://github.com/Pelagicore"
+
 # We have a recipes directory, add to BBFILES
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
 


tobsan:
It would be nice with some motivation on when this would be needed. I don't see PELUX's infrastructure (i.e. GitHub) changing any time soon, and we currently have no recipes in `meta-pelux` that point to code stored at the Pelagicore GitHub. If anything (and I'm against that too) we could add it to `meta-bistro`.

mardy:
I guess that since `meta-pelux` and `meta-bistro` are both in github.com/Pelagicore, they will both move if we decide to change hosting; and given that `meta-bistro` has actually a few recipes using a `SRC_URI` pointing to Pelagicore's github, it makes more sense to have this trick in that project, rather than here in `meta-pelux`.

Now, why would you object to having this change in `meta-bistro`? Changing repositories is not that unlikely (I remember when out set of projects moved from gitorious.org, then to code.google.com and finally to GitLab), it can happen for a multitude of reasons: github could change its policy and start charging money for corporations, or Luxoft might want to change "Pelagicore" to "Luxoft" everywhere, or we might decide that GitLab is a better place for us (it is!). :-)

breakreturn:
PELUX Baseline is also the reference for projects using this. And in many of these projects the infrastructure and network has a tendency to change over time.

tobsan:
If projects that build on PELUX Baseline has this need then sure, add it there (in meta-project-whatever).

And I disagree with the likelyhood of repo hosting changing, which I see as low. Having the full URL in the recipes makes them super clear for everyone who reads them, which imho is more valuable than the cost of a one-time search/replace action needed if we ever change hosting.

breakreturn:
@tobsan My point is that the PELUX Baseline is a reference on how to do things. That's a use-case equally valid as any other and needs to be included in the reasoning. Sure the risk of switching hosting is low in the particular case of meta-pelux or meta-bistro recipes, but so is the risk of people not understanding where they should look for the repo.

A middle ground could be to have a how-to and some rationale only in the SWF Blueprint and then let the projects needing a reference rely on that. I think in practice, people will use the PELUX layers as they are as reference though and read that part of the SWF too late.

tobsan:
@breakreturn I just see a risk of us making the PELUX Baseline so super-modular so that it becomes hard to use by itself. 

It might not sound like it, but I think the idea of having a project specific variable pointing to the project's git server is a good idea. This is similar to how Qt does it in their recipes, where SRC_URI is something like `${QT_SERVER}/${PN}.git`. But I think that makes more sense for those cases. Perhaps that's what @breakreturn means with the middle ground. Since it is a good practice when needed, I'm all for documenting this and using it where applicable.

breakreturn:
@tobsan If the Baseline would become difficult to use, we've also failed in some area ofc. But do you really think it would be more difficult to use if we have a variable like this?

Btw, I'm not as familiar with the Qt specifics to understand why the variable makes sense in their case and not ours.

tobsan:
The way I see it is that this variable makes sense for PELUX in general when we have recipes that point to stuff hosted by us. We currently have 6 regular recipes and 4 CommonAPI-recipes in `meta-bistro`, and not a single such recipe in `meta-pelux`, that point to the Pelagicore GitHub. Having to resolv a variable to find where the recipe gets its sources makes it more difficult to use, almost by definition.

The Qt layer `meta-qt5` has lots of recipes where they all follow the same release etc, so in their case all recipes use the same settings, defined in [qt5-git.inc](https://github.com/meta-qt5/meta-qt5/blob/master/recipes-qt/qt5/qt5-git.inc).

If we have projects using PELUX where it is useful to do this, we should place that logic in the project, but we can of course have it described in the SWF. The downside with that approach is that we're side-stepping the dog-fooding approach. On the other hand I suspect our SWF BluePrint will end up containing quite a lot of things that are relevant only in specific cases.

breakreturn:
@tobsan Ok, let's skip this in PELUX then.

@mardy Could you write something in the SWF Blueprint about this pattern and practice? I think the sections that describe e.g. how to create your own layer etc (I guess the PDK parts) should link to an instruction like that.

breakreturn:
@mardy and @tobsan I guess we agree to close this PR then?

mardy:
Fine with me, closing.

PR #60, opened 2017-10-27
rpi-config: Enable UART

jonte:
This allows connecting a serial console to the serial pins of the
raspberry pi.

Signed-off-by: Jonatan Pålsson <jonatan.palsson@pelagicore.com>


jonte:
I believe it is currently a default option in meta-raspberrypi, but I think the user should still verify the value of this parameter when wanting to enable UART.

PR #62, opened 2017-11-03
add short README for each variant

tobsan:

### DIFF
--- a/conf/variant/intel/README
+++ b/conf/variant/intel/README
@@ -1,0 +1,1 @@
+Variant configuration for building a non-graphical PELUX for the Intel NUC.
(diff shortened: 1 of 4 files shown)


conf/variant/intel/README
```diff
@@ -0,0 +1 @@
+Variant configuration for building PELUX for the Intel NUC without Qt+QtAuto
```
DunderRoffe: Write headless or something instead of without Qt+QtAuto. Otherwise it wont scale when adding additional variants

breakreturn: Non-graphical is used to distinguish this in a project I'm in. "Headless" means very different things to people it turns out :)

tobsan:
@DunderRoffe @breakreturn updated now.

conf/variant/rpi/README
```diff
@@ -0,0 +1 @@
+Variant configuration for building a non-graphical PELUX for the Raspberry Pi 3.
```
GordanM: For consistency reasons all README files should end with a ".".

PR #64, opened 2017-11-17
Update Pyro to align with master

thenor:


GordanM:
@thenor I am just wondering if you need to do a merge commit for that? Isn't it just a matter of pulling the master HEAD to pyro branch? I am not sure if the way it is done here might end up with different commits. @tobsan do you have more experience with this?

PR #69, opened 2017-11-27
fix for qmake in sdk

DunderRoffe:
The purpose of this PR is to add Qt tools to the SDK produced when running
```bash
bitbake -c populate_sdk core-image-peulx-qtauto-neptune
```

This requrires fixing a conflict in meta-data produced by cmake and qmake when building
the rpms for the SDK. 

The fix is setting `DIRFILES = "1"`. However this only work in `local.conf` and in distro `.conf` but not in `cmake_%.bbappend`

### DIFF
--- a/conf/variant/common/local.conf
+++ b/conf/variant/common/local.conf
@@ -10,3 +10,6 @@ PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count()}"
 PACKAGE_CLASSES ?= "package_rpm"
 
 BB_DANGLINGAPPENDS_WARNONLY = "1"
+
+# Fix for rpm metadata clash between nativesdk-cmake and nativesdk-qtbase-tools
+DIRFILES_pn-nativesdk-cmake = "1"
--- a/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
+++ b/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
@@ -6,6 +6,7 @@
 DESCRIPTION = "Reference PELUX image with QtAuto frontend"
 
 inherit core-image-pelux-qtauto
+inherit populate_sdk_qt5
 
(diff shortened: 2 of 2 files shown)


GordanM:
@DunderRoffe nice catch! Now i am wondering if this is something that will help @tobsan 's SDK commit?

DunderRoffe:
No this is about getting qmake in the SDK. @tobsan is working on build issues on the minimal SDK

Gagi2k:
"qt" is fine, just don't use "QT"

DunderRoffe:
@GordanM fixed

DunderRoffe:
updated description

conf/distro/pelux.conf
```diff
@@ -42,3 +42,6 @@ DISTRO_FEATURES_append = " systemd"
 DISTRO_FEATURES_BACKFILL_CONSIDERED += "sysvinit"
 VIRTUAL-RUNTIME_init_manager = "systemd"
 VIRTUAL-RUNTIME_initscripts = "systemd-compat-units"
+
+# Fix for rpm metadata clash between nativesdk-cmake and nativesdk-qtbase-tools
+DIRFILES_pn-nativesdk-cmake = "1"
```
tobsan: Is it possible to put this somewhere other than the distro config? Such as the image bbclass, or local.conf? If not I guess this works, but it doesn't really fit in.

DunderRoffe: As said in the description local.conf works. Did not try the bbclass could try that if you think it is better.

DunderRoffe: bbclass does not work. Shall I move it to `common/local.conf`?

tobsan: If it works to have it in the bbclass, that would make more sense imho.

tobsan: Yes please!

DunderRoffe: fixed

PR #71, opened 2017-11-29
core-image-pelux uses debug-tweaks + restructuring

tobsan:


classes/core-image-pelux.bbclass
```diff
@@ -4,26 +4,37 @@
 #
 
 inherit core-image
+inherit extrausers
+
+IMAGE_FEATURES += "ssh-server-openssh package-management"
```
Gagi2k: Does that build for you ? Especially the package-management didn't build for me without also adding IMAGE_FEATURES += ${DISTRO_FEATURES_LIBC}

DunderRoffe: It would be nice if we make sure to use the same ssh-server as we do in dev images which I belive is drop-bear?

classes/core-image-pelux.bbclass
```diff
@@ -4,26 +4,37 @@
...
+"
+
+# Default user settings
+EXTRA_USERS_PARAMS_PELUX = "\
+    usermod -P root root;\
+"
+
+# But we only set the root pw if we're not having debug-tweaks or empty-root-password in the image
+# This check is similar to the one done internally in yocto.
+EXTRA_USERS_PARAMS = "${@bb.utils.contains_any('IMAGE_FEATURES', [ 'debug-tweaks', 'empty-root-password'], '', '${EXTRA_USERS_PARAMS_PELUX}', d)}"
```
DunderRoffe: Will this give the same behavior for login in between non-dev and dev images? It would be nice if they were the same so we have a defined way of doing things :)

tobsan: Actually, it does not give the same behaviour. This PR makes non-dev images have "root" as root pw, while the -dev images will have no root password.

Not sure we really want empty root password for non-dev, do we? It would sure make it harder to build upon PELUX.

classes/core-image-pelux.bbclass
tobsan: @Gagi2k yes, this is not a change from before actually (just moving the lines around). Also the poky include sets `DISTRO_FEATURES ?= "... ${DISTRO_FEATURES_LIBC} ..."`

@DunderRoffe I wouldn't mind that at all actually. Either to use dropbear in non-dev or to use openssh in -dev images.

classes/core-image-pelux.bbclass
DunderRoffe: I agree we do not want empty root pwds for non-dev but maybe we should not have it for dev either then? It is nice if it is the same

tobsan: Yes, having "root" as the default would actually make sense to me (and it would make this code much neater). One can still override it of course.

What do you think, @breakreturn @GordanM ?

tobsan: Ping @jonte @e8johan

classes/core-image-pelux.bbclass
jonte: The non-debug image shouldn't contain an SSH server at all, right? Since you won't ever log in to it via SSH...

classes/core-image-pelux.bbclass
jonte: This is a bad idea :) On a production system, you would not have login capabilities for the root user.

breakreturn: I'd premier convenience for the user here. Anyone wanting "real project security" etc. will probably be doing many things different anyway and thus that's more of the project specific platform use-case.

GordanM: @tobsan the production image should not have a ssh-server. At least, those were the requirements from our previous experiences. I believe, @jonte presented that opinion in another comment.

jonte: > Not sure we really want empty root password for non-dev, do we? It would sure make it harder to build upon PELUX.

@tobsan What does this mean?

tobsan: I suppose this all depends on whether or not we consider `core-image-pelux` to be what we want to use in a production setting. I suspect (like @breakreturn said) that a production project would do many things differently.

However, if we want to make it easy for the user, we should just tell them to used the `-dev` image instead of the non-dev one. I'm not sure what I would prefer.

@jonte setting an empty root password for the non-dev image would mean the user of a non-dev image would have to change it (again) if they wanted the password to be disabled instead.

tobsan: @GordanM I agree, ssh server should be present only in dev image I guess.

jonte: @tobsan Hmm.. I don't really see why having "root" as the password is preferable over having an empty password (as in, no password is needed for login). It seems slightly more convenient to not have to enter "root" in the password box on each ssh command.. We're not gaining any security by having a very simple password anyway. I might be missing something?

re: disabling the login completely, this can be done for both a very simple password and an empty password.

breakreturn: As a side note, I assume that when @GordanM mentions "the production image", it is more about naming the image which is not-dev-image, right? AFAIK we are not aiming for PELUX Baseline to be production ready in the sense that it can be put in a car without project specific modifications in some areas.

classes/core-image-pelux.bbclass
tobsan: No, that makes sense. I'll remove ssh from the `IMAGE_FEATURES` for non-dev.

classes/core-image-pelux.bbclass
GordanM: @breakreturn that is correct. My thought was that PELUX should showcase this method in the Platform Baseline. This way the users would follow the same pattern for their projects. Hope this makes sense.

tobsan: Ok so it sounds like perhaps we want the non-dev image to be closer to a production ready image, but then tell our users to use the -dev image if they need things like root login and ssh access? Could this be a way forward?

tobsan: Ok, so three things:
1. The non-dev image is not a production image, so we shouldn't refer to it as such. It's simply the -dev image but without extra tools for debugging etc.
2. We want the same login behaviour for both images
3. That means we can use the empty-password tweak of yocto for the base image.

breakreturn: @tobsan Isn't there some place in the Baseline SWF where this could be stated in a nice way? I.e. a section describing the images and their purposes and considerations and rationales around those? If so, the names for the images and the reasoning of non-production-readyness could be elaborated there.

tobsan:
@breakreturn I actually added a short text about the differences in dev and non-dev images, but I'll update that after this is merged.

breakreturn:
@tobsan Ok great.

PR #77, opened 2017-12-19
README.md: Only support two latest yocto releases

thenor:


tobsan:
Since pyro should be aligned with master now, I don't think we need this PR.

thenor:
Pyro will instead be updated with this change by merging master into the branch.

PR #81, opened 2018-01-02
systemd-networkd: add default DHCP for ethernet

tobsan:
Will bring up any ethernet device by default by using DHCP, and also
sets up resolv.conf in a way so that both systemd and non-systemd aware
applications will work.

### DIFF
--- a/recipes-core/systemd/systemd/10-pelux-ethernet.network
+++ b/recipes-core/systemd/systemd/10-pelux-ethernet.network
@@ -1,0 +1,6 @@
+[Match]
+Name=en*
+Name=eth*
+
+[Network]
+DHCP=yes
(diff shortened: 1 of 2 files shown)


recipes-core/systemd/systemd/10-pelux-ethernet.network
```diff
@@ -0,0 +1,6 @@
+[Match]
+Name=en*
+Name=eth*
```
GordanM: I guess this has been the case on all the devices we are trying to support? I can't remember any other common pattern for the naming of the ETH iface.

tobsan: Yes, the naming is taken from how they are named on the NUC and the Pi. I suppose we can amend this list later on if needed.

GordanM: Good!

PR #83, opened 2018-01-05
Bump version of Neptune

tobsan:
Fixes an issue where Neptune would crash immediately at startup.


tobsan:
@GordanM actually this bumps Neptune to work the same way in morty as it does in pyro.

PR #84, opened 2018-01-10
Add configurable static ip address to the image

BassemMohsen:
Configuring STATIC_IP_ADDRESS fixes the issue with poky ssh tests.

### DIFF
--- a/conf/variant/common/local.conf
+++ b/conf/variant/common/local.conf
@@ -4,6 +4,10 @@ DL_DIR = "${TOPDIR}/downloads"
 SDKMACHINE = "x86_64"
 DISTRO = "pelux"
 
+# Target Static IP address, Override this to configure a static
+# ip address for development purposes such as poky ssh and ping test.
+STATIC_IP_ADDRESS = ""
+
 BB_NUMBER_THREADS ?= "${@oe.utils.cpu_count()}"
 PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count()}"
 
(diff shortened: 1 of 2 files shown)


recipes-core/systemd/systemd_%.bbappend
```diff
@@ -7,5 +7,9 @@ SRC_URI += "file://10-pelux-ethernet.network"
 # We also make sure resolv.conf is using systemd-resolved
 do_install_append() {
     install -m 0644 ${WORKDIR}/10-pelux-ethernet.network ${D}${sysconfdir}/systemd/network/10-pelux-ethernet.network
+    if [ ! -z "${STATIC_IP_ADDRESS}" ]; then
+    	echo '[Address]' >> ${D}${sysconfdir}/systemd/network/10-pelux-ethernet.network
+    	echo "Address=${STATIC_IP_ADDRESS}/24" >> ${D}${sysconfdir}/systemd/network/10-pelux-ethernet.network
```
tobsan: Make up your mind about single quotes versus double quotes.

conf/variant/common/local.conf
```diff
@@ -4,6 +4,9 @@ DL_DIR = "${TOPDIR}/downloads"
 SDKMACHINE = "x86_64"
 DISTRO = "pelux"
 
+# Target Static IP address
```
tobsan: A better comment would be nice here. Why is this a good thing, what is it used for etc.

BassemMohsen: I added the comment now.

recipes-core/systemd/systemd_%.bbappend
BassemMohsen: double quotes usage is unified.

PR #85, opened 2018-01-11
qemu variant: Fix broken include path

sergeevaeka:


sergeevaeka:
@tobsan please review

conf/variant/qemu-x86-64_nogfx/bblayers.conf.sample
```diff
@@ -1,4 +1,8 @@
-# This is needed for bitbake to pick up the includes
+# These are needed in order to make bitbake pick up the includes
 BSPDIR := "${TOPDIR}/.."
+BBPATH := "${TOPDIR}:${BSPDIR}/sources/meta-pelux"
 
 require conf/variant/common/bblayers.conf
+
+BBLAYERS += "\
```
tobsan: Is this really needed though?

jonte: Probably not - we left it in there since it was present in a different variant file (from a different project). Would you rather see it removed?

tobsan: Yes, if it doesn't add anything beside what's in the common `bblayers.conf`, please remove it.

PR #86, opened 2018-01-22
Update meta-pelux to rocko

tobsan:
This adds some changes needed to adapt to the rocko release of poky for meta-pelux.

### DIFF
--- a/conf/variant/common/bblayers.conf
+++ b/conf/variant/common/bblayers.conf
@@ -3,12 +3,11 @@ LCONF_VERSION = "7"
 BBFILES  ?= ""
 BBLAYERS ?= "                                         \
   ${BSPDIR}/sources/poky/meta                         \
-  ${BSPDIR}/sources/poky/meta-yocto                   \
+  ${BSPDIR}/sources/poky/meta-poky                    \
   ${BSPDIR}/sources/poky/meta-yocto-bsp               \
   ${BSPDIR}/sources/meta-openembedded/meta-oe         \
   ${BSPDIR}/sources/meta-openembedded/meta-networking \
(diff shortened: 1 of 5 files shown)


conf/variant/common/bblayers.conf
```diff
@@ -3,12 +3,11 @@ LCONF_VERSION = "7"
 BBFILES  ?= ""
 BBLAYERS ?= "                                         \
   ${BSPDIR}/sources/poky/meta                         \
-  ${BSPDIR}/sources/poky/meta-yocto                   \
+  ${BSPDIR}/sources/poky/meta-poky                    \
   ${BSPDIR}/sources/poky/meta-yocto-bsp               \
   ${BSPDIR}/sources/meta-openembedded/meta-oe         \
   ${BSPDIR}/sources/meta-openembedded/meta-networking \
   ${BSPDIR}/sources/meta-openembedded/meta-python     \
-  ${BSPDIR}/sources/meta-openembedded/meta-ruby       \
```
GordanM: I'm glad we are getting rid of this but is there an explanation why?

tobsan: The explanation is that `meta-ruby` contained very few recipes, so those that were there were moved somewhere else and the layer was deleted.

GordanM: Okay, perfect so it's not us stopping to use it but the layer content has been moved to other places. Could that be reflected in the commit message or do you think it is redundant?

tobsan: No, it should be mentioned. Will update.

tobsan: Done

GordanM:
Thanks, I'm good!

PR #87, opened 2018-01-24
packagegroup: Change to use debug-utils for dev images

thenor:


tobsan:
Looks good. Let's wait until it's merged in bistro as well.

thenor:
@tobsan Yes, for that reason I could set it to WIP.

PR #88, opened 2018-01-26
variant/common: Don't install extra devel deps for QEMU

jonte:


conf/variant/common/local.conf
```diff
@@ -17,3 +17,6 @@ BB_DANGLINGAPPENDS_WARNONLY = "1"
 
 # Fix for rpm metadata clash between nativesdk-cmake and nativesdk-qtbase-tools
 DIRFILES_pn-nativesdk-cmake = "1"
+
+# Don't install any extra devel packages just because we're running on QEMU
```
thenor: I'm assuming "devel" is short for "development", but why not at least write out the word in the comment?
If there some (official) background reason to this decision could you please include it in the commit message (not on the first line).

jonte: I just tried staing < 80 chars in width ;)

jonte:
@tobsan I updated the commit message to further motivate the change

@thenor -s added!

PR #91, opened 2018-02-08
Swupdate PoC for the Raspberry Pi 3

FlorentRevest:
This series allowed me to generate a basic SWUpdate bundle from Yocto and install it from PELUX on a Raspberry Pi 3. It works using an A/B update scheme.

SWUpdate being an "update framework" and not a standalone solution, it needs a fair bit of integration work. One of the first requirements of SWUpdate for an A/B update scheme is to use a configurable U-Boot. Hence, the first few patches compile PELUX with U-Boot support on the Raspbery Pi 3. It also prepares the terrain so that u-boot-fw-utils can be used by SWUpdate to set U-Boot envvars and dynamically switch between the two available rootfs.

Then, SWUpdate is configured without unneeded options such as the integrated http updates upload server.

Finally, a core-image-pelux-minimal-update recipe is added. It is able to generate a .swu update based on a core-image-pelux-minimal compressed image.

Any comments are welcome!

This series is loosely based on https://github.com/sbabic/meta-swupdate-boards (with various fixes applied) which is only meant to serve as an example and not integrated directly to pelux-manifests.


recipes-bsp/u-boot/u-boot-fw-utils_%.bbappend
```diff
@@ -0,0 +1,10 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:${THISDIR}/files:"
```
tobsan: Do you really need both here?

recipes-extended/images/core-image-pelux-minimal-update/raspberrypi3/emmcsetup.lua
```diff
@@ -0,0 +1,85 @@
...
+	local s1
+	local ret
+
+	local log = os.tmpname()
+
+	local eMMC = "/dev/mmcblk0"
+	ret = file_exists("/dev/mmcblk0")
+
+	if (ret == false) then
+		return false, "Cannot fine eMMC"
```
tobsan: Should be "find" I think.

recipes-support/swupdate/swupdate/raspberrypi3/swupdate
```diff
@@ -0,0 +1,191 @@
+#! /bin/sh
```
tobsan: Will this play nicely with the systemd setup we have? I don't remember if systemd generates a wrapper for such scripts or not.

FlorentRevest:
u-boot-fw-utils_%.bbappend: indeed, we didn't need both, good catch! :)
emmcsetup.lua: fixed!
swupdate: actually we don't need to override this script yet. there was just a couple of adaptations from the default init script to run swupdate in "suricatta" mode with a correct option depending on the current rootfs. for this simple proof of concept, it's not needed. I removed the file for now and when we will really need it I'll make sure we have a proper systemd service.

FlorentRevest:
I updated the PR with most of the rpi-specific files moved to meta-rpi-extras. (I forgot that part in my previous update)
However, I chose to keep some raspberrypi3 files in _recipes-extended/images/core-image-pelux-minimal-update/raspberrypi3/_ (in the last commit) because I really think it makes sense to keep the .swu descriptions centralized in one place.
I also added the missing Signed-off-by.

recipes-extended/images/core-image-pelux-minimal-update/raspberrypi3/emmcsetup.lua
```diff
@@ -0,0 +1,85 @@
...
+	cmdexec("/usr/sbin/partprobe " .. eMMC)
+
+	return true, out
+end
+
+function postinst()
+	local out = "Post installed script called"
+
+	return true, out
+end
```
tobsan: Is there a specific reason why this is written in lua?

FlorentRevest: SWUpdate handlers can be scripted in Lua: https://sbabic.github.io/swupdate/handlers.html#lua-handlers
This is also what they use in their reference integration layer: https://github.com/sbabic/meta-swupdate-boards/blob/master/recipes-extended/images/update-image/raspberrypi3/emmcsetup.lua

PR #97, opened 2018-03-08
Add QEMU launch script

chuanjin:
Signed-off-by: Chuan Jin <cjin@luxoft.com>

### DIFF
--- a/scripts/qemu_launcher.sh
+++ b/scripts/qemu_launcher.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+
+# Copyright (C) 2018 Luxoft
+#
+# This is a handy script used to launch QEMU easily.
+# It wraps up the kvm command and provides simple options to lanuch dev and prod images
+# without loosing the flexibility to adjust parameters like memory, cpu, etc.
+#
+# To see the usage, run:
+# ./qemu_launcher.sh -h
+#
+# e.g.
+# To start dev image with static IP (asumming the current folder is the build)
+# ./qemu_launcher.sh -d -s
+#
+#
+
+command -v kvm > /dev/null
+if [ $? != 0 ]
+then
+    echo "Warning:"
+    echo "kvm is not installed yet, please install it with command below and rerun the script"
+    echo "sudo apt-get install qemu-kvm libvirt-clients libvirt-daemon-system "
+    exit 1
+fi
+
+set -e
+
+# Print usage information of the script
+usage() {
+    echo "usage: $0 [options]"
+    echo -e "[options] is any of the following:"
+    echo -e "  -f | --folder           \tSpecify a source folder of the QEMU image"
+    echo -e "                          \tDefaults to tmp/deploy/images/qemux86-64"
+    echo -e "  -i | --hda              \tSpecify rootfs file. Defaults to core-image-pelux-minimal-qemux86-64.ext4"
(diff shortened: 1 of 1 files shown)


chuanjin:
Please confirm the header of the script, should we use Luxoft or Pelagicore here?
A minor change is I rename the default image from "core-image-ntg-x-qemux86-64.ext4"  to "core-image-pelux-minimal-qemux86-64.ext4", same for dev image.

scripts/qemu_launcher.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
+
+# Copyright (C) 2017-2018 Luxoft
```
GordanM: To me this is OK. @jeremiah could you please verify?

jeremiah: Fine with me. :-) 

As a side note however, copyright lasts for 70+ years past the life of the author. So we can just put one year in there, we don't have to put a range. This is trivial and pedantic of me to me to even mention, but there it is.

scripts/qemu_launcher.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
+
+# Copyright (C) 2018 Luxoft
+# All rights reserved
+#
+# This is a handy script used to launch QEMU with different options easily
```
thenor: What does this sentence mean? I don't understand the "different options easily" part. Does it mean that one can add options on command line which are sent to the command? I think this sentence would benefit from being split up into two or more sentences for readability.

thenor: I think I understand from reading the code, but I still think that it would be nice from a readability point to rephrase and expand the description. Otherwise the description becomes unnecessary, but I think a description like this is a good introduction, so just expand it slightly.

chuanjin: Sorry for the confusion. I add a few more sentences to explain a bit more.

jeremiah:
I think using Luxoft as the copyright holder makes the most sense. If I may I’d like to ask that we remove the line “All rights reserved” because we don’t actually reserve all our rights, we’re granting the rights to modify, study, and use.

thenor:
I'll add @tobsan as a reviewer just as an FYI about the image renaming. He won't have time to review since he's on vacation.

scripts/qemu_launcher.sh
```diff
@@ -0,0 +1,159 @@
...
+    fi
+}
+
+setup_dev() {
+    DEV=true
+    if [ $hda == "core-image-pelux-minimal-qemux86-64.ext4"  ]
+    then
+        hda="core-image-pelux-minimal-dev-qemux86-64.ext4"
+    fi
+}
```
tobsan: Sorry but what does this code do? If hda is set, we set it again to the same value?

scripts/qemu_launcher.sh
```diff
@@ -0,0 +1,159 @@
...
+    echo -e "                          \tDefaults to core-image-pelux-minimal-dev-qemux86-64.ext4"
+    echo -e "  -s | --static           \tRun QEMU development image with static IP (192.168.7.2)"
+    echo -e "  -l | --list             \tList all available images from the default folder or folder user specified"
+    echo -e "  -h | --help             \tDisplay this help"
+    exit 0
+}
+
+list_images() {
+    echo "Available images:"
+    ls $folder/*.ext4 | xargs -n 1 basename
```
tobsan: Are .ext4 the only format accepted by QEMU?

PR #100, opened 2018-03-16
core-image-pelux: append swupdate only for rpi

thenor:
This commit is needed to be able to run the smoke tests.

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -26,7 +26,7 @@ IMAGE_INSTALL += "\
 "
 
 # OTA mechanism
-IMAGE_INSTALL += "\
+IMAGE_INSTALL_append_rpi += "\
     swupdate \
 "
 


FlorentRevest:
Fine with me. I'll remember this if we ever need to include swupdate on qemu builds.

recipes-core/images/core-image-pelux-minimal.bb
```diff
@@ -6,3 +6,4 @@
 DESCRIPTION = "Image for creating a Pelux image"
 
 inherit core-image-pelux
+IMAGE_INSTALL_remove_qemux86-64 = "template-service swupdate"
```
GordanM: It would be interesting to see why is the `template-service` a problem for the testimage builds? Is there any explanation. `swupdate` is quite big so I think it might affect the image but the `template-service` is nothing more than a `int main()` C++ executable.

thenor:
@FlorentRevest: I only needed it to be available for testing purposes. I think we might want it for running the smoke tests, but we might be able to do it some other way.
@GordanM: @erikboto wrote to me saying that we don't actually need to exclude the template service. It was a copy-paste-error from his side (I have gotten help from him with this change). So it might be unnecessary to exclude it.

thenor:
@tobsan If we have a Jenkinsfile in pelux-manifests for smoke testing in qemu, can we also have a tests image where we exclude things like this? It feels it might be unnecessary to exclude it always for qemu? Or how do you think in this situation?

tobsan:
I'm not sure how the test image stuff works, but one can probably do lots of customizations when building it, since it is a yocto target.

When it comes to swupdate, perhaps it would make more sense to include it ONLY in the rpi build as `IMAGE_INSTALL_append_rpi = "swupdate"` or something like that. That way the problem will go away on its own. What do you think @FlorentRevest

FlorentRevest:
@tobsan it totally makes sense to include swupdate only on supported platforms. I don't mind at all.

thenor:
Then I can change this to what @tobsan suggested. Append instead of remove.

PR #101, opened 2018-03-19
Add powerbutton script as append to acpi

tobsan:
We want the NUC to respond to power button presses, this is done by
adding an acpi rule with a corresponding script as action, causing the
system to shutdown on press.


tobsan:
Just FYI: I tested this manually and through a Yocto build and it works like a charm!

tobsan:
@GordanM yeah it would. However I believe that since the event matching is for anything called a powerbutton, it would make sense for all of those anyway.

PR #102, opened 2018-03-19
Make static IP configurable

chuanjin:
Signed-off-by: Chuan Jin <cjin@luxoft.com>

### DIFF
--- a/scripts/qemu_launcher
+++ b/scripts/qemu_launcher
@@ -45,7 +45,11 @@ usage() {
     echo -e "                          \tDefaults to 3333"
     echo -e "  -d | --dev              \tRun QEMU development image instead of normal image"
     echo -e "                          \tDefaults to core-image-pelux-minimal-dev-qemux86-64.ext4"
-    echo -e "  -s | --static           \tRun QEMU development image with static IP (192.168.7.2)"
+    echo -e "  -s | --static           \tThe switch to toggle running QEMU dev image with static IP (Defaults to false)"
+    echo -e "  -I | --target_ip        \tSpecify the target static IP, must be 192.168.7.x (Defaults to 192.168.7.2)"
+    echo -e "                          \tPlease aware that to change the default static IP, it is also needed to update the"
+    echo -e "                          \tSTATIC_IP_ADDRESS in local.conf in meta layer and rebuild the image"
+    echo -e "  -g | --server_ip        \tSpecify the gateway server IP, must be 192.168.7.x (Defaults to 192.168.7.1)"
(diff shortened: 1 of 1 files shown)


scripts/qemu_launcher
```diff
@@ -153,7 +163,7 @@ else
     then
         start_qemu "-net nic"
     else
-        cmdline="$cmdline ip=192.168.7.2::192.168.7.1:255.255.255.0"
-        start_qemu "-device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:02 -netdev tap,id=net0,ifname=tap0,script=no,downscript=no"
+        cmdline="$cmdline ip=$target_ip::$server_ip:255.255.255.0"
```
tobsan: Actually, the netmask should be configurable as well, shouldn't it? I guess it depends on how one sets up the tap device though, so maybe not? @BassemMohsen what do you think?

BassemMohsen: the script runqemu has hardcoded the subnet mask, you can look at line 946 here:
https://git.yoctoproject.org/cgit.cgi/poky/plain/scripts/runqemu

I guess it's not the optional for yocto ssh and ping tests.

chuanjin: So the static IP and gateway should be better like 192.168.7.x maybe?

BassemMohsen: Also it's a long shot the user would be interested to use subnet masks here. and usually 24 mask is what you use with ip4.

tobsan: If the netmask is hardcoded upstream anyway then nevermind :)

scripts/qemu_launcher
```diff
@@ -45,7 +45,11 @@ usage() {
     echo -e "                          \tDefaults to 3333"
     echo -e "  -d | --dev              \tRun QEMU development image instead of normal image"
     echo -e "                          \tDefaults to core-image-pelux-minimal-dev-qemux86-64.ext4"
-    echo -e "  -s | --static           \tRun QEMU development image with static IP (192.168.7.2)"
+    echo -e "  -s | --static           \tThe switch to toggle running QEMU dev image with static IP (Defaults to false)"
+    echo -e "  -I | --target_ip        \tSpecify the target static IP (Defaults to 192.168.7.2)"
+    echo -e "                          \tPlease aware that to change the default static IP, it is also needed to update the"
+    echo -e "                          \tSTATIC_IP_ADDRESS in conf/variant/common/local.conf in meta layer and rebuild the image"
```
tobsan: I think it's better to just say update `local.conf` without stating location of that file (it's always in <builddir>/conf/ anyway)

PR #103, opened 2018-03-30
README: update copyright years

sashko:


jeremiah:
Copyright doesn't need to be updated. It lasts for the life of the author +70 years.

sashko:
@jeremiah yes, I know. Thanks for the comment. It is usually used to indicate that the project is active.

PR #104, opened 2018-04-03
Excluding output image formats .hddimage, .iso, and .tar.bz2 from dep…

SKumarDas:
Deprecating ..hddimage, .iso, and .tar.bz2 image formats as we don't need it. 
Please help to review and merge it.

Regards,
Sapan Das


conf/variant/common/local.conf
```diff
@@ -20,3 +20,7 @@ DIRFILES_pn-nativesdk-cmake = "1"
 
 # Don't install any extra devel packages just because we're running on QEMU
 POKYQEMUDEPS_remove = "packagegroup-core-device-devel"
+
+# Don't build .hddimg and .iso image formats as we don't need it.
```
GordanM: No need for the `.` at the end of the line here. And s/it/them

tobsan: We are using `.rpi-sdimg` and `.wic` for pi and NUC, respectively.

classes/core-image-pelux.bbclass
```diff
@@ -40,3 +40,4 @@ toolchain_create_sdk_env_script_append() {
 IMAGE_ROOTFS_SIZE ?= "1000000"
 IMAGE_FSTYPES ?= "ext3 sdcard"
 IMAGE_FSTYPES += "ext3.gz"
```
tobsan: Do we need this filetype at all?

breakreturn:
Since we are in some way providing experimental support for TX-2, or at least have the ambition (the last time I checked) to do so, wouldn't we take a step away from that with this? While also breaking things for current users of TX-2?

Qemu isn't affected by the change either I guess?

GordanM:
@breakreturn good point! This should be taken into account. Could you please connect with someone from their team @SKumarDas ?

conf/variant/common/local.conf
```diff
@@ -20,3 +20,7 @@ DIRFILES_pn-nativesdk-cmake = "1"
...
+NOHDD = "1"
+NOISO = "1"
```
erikboto: Can we use the more override-friendly version ?= to assign NOHDD and NOISO here?

classes/core-image-pelux.bbclass
```diff
@@ -40,3 +40,4 @@ toolchain_create_sdk_env_script_append() {
...
+IMAGE_FSTYPES_remove = "tar.bz2"
```
erikboto: Is the assignment on line 41 not happening? Seems weird that we would need this _remove line if the assignment is happening. And if it's not happening, it should probably be removed for clarity or changed to a non-conditional assignment to force it.

Also line 42 seems weird if we want ext3.gz it should just be added in the line 41.

tobsan: Isn't the assignment on 41 a strong default, so it may be overridden by, for example, machine configurations? I agree that we should make this clearer though.

erikboto: Yes it can be overridden, which is good. But why are we then explicitly adding ext3.gz and removing tar.bz2 afterwards? What if someone wants tar.bz2? Shouldn't we either set our preferred defaults using ?=, and then not touch if if someone chooses to override it?

conf/variant/common/local.conf
erikboto: I guess it can still be overridden in a machine specific local.conf if it's done after including common/local.conf, so I guess it's ok to keep the assigment as it is now.

classes/core-image-pelux.bbclass
FlorentRevest: I need the ext3.gz for SWUpdate artifacts

SKumarDas: @FlorentRevest, is it is okay to remove .tar.bz2? Or you will need this image format as well?

FlorentRevest: @SKumarDas you can remove it, I don't need that one

SKumarDas: @breakreturn Could you please conform what image format you use in qemo? Do you use tar.bz2 in any case?

breakreturn: @SKumarDas Maybe @jonte or @erikboto can support with this.

conf/variant/common/local.conf
```diff
@@ -20,3 +20,10 @@ DIRFILES_pn-nativesdk-cmake = "1"
...
+
+# Removing tar.bz2 image format as it is not used. Whoever needs it, they can add it in respective variant's local.conf file
```
tobsan: Perhaps we can reword this as: whoever needs this will supply it in their own ``local.conf``. The variant concept is something relatively local to us.

PR #107, opened 2018-04-09
Introduced QtAS packagegroups in PELUX

TSAnsari:
b2qt defines some packagegroups for QtAS addons(neptune-ui is one of the addons) as
well as packagegroups for host & target SDK. PELUX is now using these packagegroups
to introduce these QtAS addons & SDK instead of manually adding each component. One
consequence of using packagegroups from QtAS is that we are introducing additional
addons that were not part of PELUX distribution before.

### DIFF
--- a/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
+++ b/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
@@ -8,5 +8,5 @@ DESCRIPTION = "Reference PELUX image with QtAuto frontend"
 inherit core-image-pelux-qtauto
 inherit populate_sdk_qt5
 
-# This image uses neptune as the reference UI
-IMAGE_INSTALL += " neptune-ui "
+# This image uses neptune as the reference UI, which is one of the addons in Qt Auto packagegroup
+IMAGE_INSTALL += " packagegroup-b2qt-automotive-addons "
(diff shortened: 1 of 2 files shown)


layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
```diff
@@ -8,5 +8,5 @@ DESCRIPTION = "Reference PELUX image with QtAuto frontend"
 inherit core-image-pelux-qtauto
 inherit populate_sdk_qt5
 
-# This image uses neptune as the reference UI
-IMAGE_INSTALL += " neptune-ui "
+# This image uses neptune as the reference UI, which is one of the addons in QtAS packagegroup
```
GordanM: QtAS was discarded as a shorthand for Qt Automotive by the TQC marketing, we should use `Qt Auto` instead. The same goes for the commit message.

layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune.bb
```diff
@@ -8,5 +8,5 @@ DESCRIPTION = "Reference PELUX image with QtAuto frontend"
 inherit core-image-pelux-qtauto
 inherit populate_sdk_qt5
 
-# This image uses neptune as the reference UI
-IMAGE_INSTALL += " neptune-ui "
+# This image uses neptune as the reference UI, which is one of the addons in Qt Auto packagegroup
```
tobsan: Nice, this makes sense as it is a qtauto specific image!

PR #108, opened 2018-04-10
Add TTYs to make sure a tty is available always

SKumarDas:
Currently there is only one tty which is taken up by the UI application.
Because of which debugging becomes difficult. This code change will enable having multiple
TTYs (4 TTYs as of now).

Signed-off-by: Sapan Das <SKumarDas@luxoft.com>

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -12,6 +12,13 @@ do_install_append() {
     	echo "Address=${STATIC_IP_ADDRESS}/24" >> ${D}${sysconfdir}/systemd/network/10-pelux-ethernet.network
     fi
     ln -s /run/systemd/resolve/resolv.conf ${D}${sysconfdir}/resolv.conf
+
+    # Create mutiple consoles
+    tmp="${TTY_LIST}"
+    for ttydev in $tmp ; do
+        ln -sf ${systemd_unitdir}/system/getty@.service ${D}${sysconfdir}/systemd/system/getty.target.wants/getty@$ttydev.service
+    done
 }
 
 FILES_${PN} += "/etc/resolv.conf"
+FILES_${PN} += "${sysconfdir}/systemd/system/getty.target.wants"
--- a/conf/variant/common/local.conf
+++ b/conf/variant/common/local.conf
@@ -27,3 +27,6 @@ NOISO = "1"
 
 # Removing tar.bz2 image format as it is not used. Incase of need, it can be supplied through own local.conf
 IMAGE_FSTYPES_remove = "tar.bz2"
+
+# List of ttys
+TTY_LIST = "tty2 tty3 tty4 tty5"


GordanM:
Nice job! 

Just a small note on the commit log: it would be better to use the passive voice, instead of "we have" use "there is". Also, the 1st and the 3rd sentence of the log are enough there is no need to write a lot and lose the core of the message. This is a good excercise because when contributing to the software-factory it would be good to be concise and not add a lot of text that doesn't give value. Do you agree with this?

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -12,6 +12,12 @@ do_install_append() {
     	echo "Address=${STATIC_IP_ADDRESS}/24" >> ${D}${sysconfdir}/systemd/network/10-pelux-ethernet.network
     fi
     ln -s /run/systemd/resolve/resolv.conf ${D}${sysconfdir}/resolv.conf
+
+    # Create mutiple consoles
+    tmp="${TTY_LIST}"
+    for ttydev in $tmp ; do
```
tobsan: It should be possible to do just `for ttydev in "${TTY_LIST}"`, right?

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -12,6 +12,12 @@ do_install_append() {
...
+        ln -sf ${systemd_unitdir}/system/getty@.service ${D}${sysconfdir}/systemd/system/getty.target.wants/getty@$ttydev.service
```
tobsan: You have probably tried this, but just want to check if this is the preferred place to put those service files, as opposed to, for example, `multi-user.target.wants`.

SKumarDas: These service files will get created as part of getty.target. So whichever run level includes getty.target, These service files will be created.

recipes-core/systemd/systemd_%.bbappend
SKumarDas: To make it look clean I have done this way, as in other .bb file systemd-serialgetty.bb  they have done this way.

tobsan:
@SKumarDas you mentioned yesterday that you had issues with the symlinks not being created unless you did a fresh build, is that still the case?

recipes-core/systemd/systemd_%.bbappend
SKumarDas: @GordanM  Yes still now my changes on creating symlink do not reflect in subsequent build.

tobsan: I would try to do a ``rm -rf ${D}${sysconfdir}/systemd/system/getty.target.wants/`` before the symlink step then.

SKumarDas: @GordanM  @tobsan , The image is containing my changes after making change to creating symlink. It is just that the dd command that I use to write the image to pendrive is not writing the newly introduced symlinks. If I am formatting the pendrive and then using dd cmd to write the image, it is actually showing the newly introduced symlinks as well.

tobsan: That's strange. Can you try to add `FILES_${PN} += ${sysconfdir}/systemd/system/getty.target.wants` and see if that changes anything?

GordanM:
@SKumarDas please use the workflow stated at http://pelux.io/software-factory/master/swf-blueprint/docs/articles/workflow/git-workflow.html and address my comments before merging this PR.

tobsan:
@GordanM can you approve this?

PR #109, opened 2018-04-17
qtbase_%.bbappend: only match with end of line

tobsan:
The previous version led to double match if the same recipe was built
more than once. so python (unavailable => python3 (correct) => python33
(didn't work at all!).


GordanM:
Nice catch! Should we somehow add a comment above the changed line to note that this needs to be upstreamed?

tobsan:
@GordanM yes, we should. I'm just not sure I would have time to correct issues with the patch before I leave, so I'd rather not be the one to upstream this one.

PR #113, opened 2018-04-24
Generify SWUpdate artifacts targets

FlorentRevest:
We rely on the meta-swupdate's swupdate.bbclass to generate .swu update
artifacts for a given image. However, this bbclass requires to have one
recipe per artifact. Since we have four different targets, we need to
factorize the work between those four recipes.

This patch introduces a common pelux-image-update.inc file that can be
included by all the artifacts recipes to generate appropriate artifacts
from a set of generic config files.

This inc file finds sw-descriptor and scripts in MACHINE-specific
subdirectories of the pelux-image-update/ directory. In order to be kept
generic, those sw-desciptors can refer to the "PELUX_IMAGE" token that
will be replaced by the name of the necessary PELUX image dynamically.


recipes-extended/images/pelux-image-update/raspberrypi3/sw-description
```diff
@@ -5,7 +5,7 @@ software = {
     stable : {
         main : {
             images: ({
-                filename = "core-image-pelux-minimal-raspberrypi3.ext3.gz";
+                filename = "PELUX_IMAGE";
```
GordanM: What is the motivation behind this name change? I guess it's the python script?

FlorentRevest: Basically, this sw-description used to be specific to "core-image-pelux-minimal" target so it hardcoded the name of the generated rootfs: core-image-pelux-minimal-raspberrypi3.ext3.gz
However, since we would like to have one sw-description for every target, we can't hardcode the name of one specific .ext3.gz
This "PELUX_IMAGE" placeholder, both documented in the .inc file and in the commit message, is dynamically replaced by my python script with "core-image-pelux-minimal-raspberrypi3.ext3.gz" or "core-image-pelux-minimal-dev-raspberrypi3.ext3.gz" or "core-image-pelux-qtauto-neptune-raspberrypi3.ext3.gz" or "core-image-pelux-qtauto-neptune-dev-raspberrypi3.ext3.gz" or whatever we could need next...

GordanM: Yup got the point thanks!

FlorentRevest: I'm not saying this is a beautiful solution I am proud of but I couldn't find anything better given the constraints of meta-swupdate :smiley:

recipes-extended/images/pelux-image-update.inc
```diff
@@ -0,0 +1,57 @@
...
+    # actual name of our compressed rootfs.
+    workdir = d.getVar('WORKDIR', True)
+    swdescpath = os.path.join(workdir, "sw-description")
+
+    swdesc = ""
+    with open(swdescpath) as f:
+        swdesc = f.read()
+
+    appendmachine = d.getVarFlag("SWUPDATE_IMAGES_NOAPPEND_MACHINE", peluximg, True)
+    if appendmachine == None:
```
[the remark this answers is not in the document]
FlorentRevest: I get your point but the goal here was to reproduce exactly the way meta-swupdate constructs the image name (to make sure we actually access the same file :) )
And they do it this way https://github.com/sbabic/meta-swupdate/blob/master/classes/swupdate.bbclass#L128

FlorentRevest:
@DunderRoffe, mhh... for whatever reason I can't directly answer your comment on "PELUX_IMAGE_NAME_PLACEHOLDER" but yes I definitely agree with you. I wanted to do it and then forgot.
Do you think it would be appropriate to surround it with fancy characters such as %%PELUX_IMAGE_NAME_PLACEHOLDER%% to make it even more obvious this is a dynamic value ?

recipes-extended/images/pelux-image-update.inc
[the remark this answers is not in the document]
DunderRoffe: Ok that sounds like a good thing :)

DunderRoffe:
Sure that might be a solution :)

FlorentRevest:
There you go :grinning:

GordanM:
@TSAnsari have you seen these build failures before? I thought that you and @tobsan encountered some issues similar to this but it was an intermiten failure right? @FlorentRevest can you retrigger the build?

GordanM:
Merging even though there is a build failure in Qt Auto GammaRay. This should be investigated separately.

PR #114, opened 2018-04-27
core-image-pelux: Include an sftp server by default

FlorentRevest:
QtCreator uses sftp deployment (instead of scp) to "Generic embedded
Linux" devices. Without including an sftp server, it fails with
"The SFTP server finished unexpectedly with exit code 127" but with that
package included, it Just Works(TM).

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -5,7 +5,7 @@
 
 inherit core-image
 
-IMAGE_FEATURES += "ssh-server-openssh package-management debug-tweaks"
+IMAGE_FEATURES += "package-management debug-tweaks"
 
 # Include softwarecontainer only if the process-containment feature has been enabled
 IMAGE_INSTALL += "\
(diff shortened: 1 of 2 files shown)


classes/core-image-pelux.bbclass
```diff
@@ -7,6 +7,11 @@ inherit core-image
 
 IMAGE_FEATURES += "ssh-server-openssh package-management debug-tweaks"
 
+# Complete ssh-server-openssh with a sftp server
+IMAGE_INSTALL += "\
+    openssh-sftp-server \
```
GordanM: Shouldn't the `openssh` features be included only in the dev image?

frznlogic: What you said ;)

frznlogic:
I think this PR is flawed, we don't need sshd in the core image. The -dev images should be used for development, not the core one. IMHO.

FlorentRevest:
I completely agree that we don't need SSH on the core image. However, sshd is *already* there. I only added support for sftp.

I could update the PR to move sshd from the core-image to the -dev image and add sftp support along the way.

GordanM:
@FlorentRevest that sounds great!

frznlogic:
Sounds much better in that case :)

FlorentRevest:
Updated ;)

PR #118, opened 2018-05-24
Add arp-driver to arp pelux image

TSAnsari:
Signed-off-by: tansari <tansari@luxoft.com>


thenor:
@frznlogic It was a renaming of the layer. I guess that could have been mentioned in the commit message.

frznlogic:
Ah ok, thanks @thenor for the update ;).

PR #119, opened 2018-05-24
README and LICENSE cleanups

sashko:


README.md
```diff
@@ -12,15 +12,17 @@ layer are specific for the PELUX image but can be used outside of this context.
...
-* meta-swupdate (https://github.com/sbabic/meta-swupdate)
+
+------------
+| meta layer             | git repo |
+| ---------------------- |----------|
+| poky                   | http://git.yoctoproject.org/cgit/cgit.cgi/poky |
+| meta-openembedded      | http://cgit.openembedded.org/meta-openembedded |
+| meta-virtualization    | http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization |
+| meta-bistro            | https://github.com/Pelagicore/meta-bistro |
+| meta-swupdate          | https://github.com/sbabic/meta-swupdate |
```
frznlogic: I like the table , but can you align everything a little bit perhaps so it reads more nicely in pure text format as well?

sashko: I am not sure I have understood your comment: everything is well aligned.

frznlogic: the top line is too short, the end git repo column doesnt align, etc;)

frznlogic: Hm, it shouldn't and the docs you point to say that either or is possible? Look at the first vs the second example, both are valid, but the first one looks much cleaner if you see it in text formatting.

PR #120, opened 2018-05-29
pelux-image-update: fix path to sw-description

sashko:
Fixed path to sw-description and added SRVREV for it to get rid of the
warnings.

### DIFF
--- a/recipes-extended/images/pelux-image-update.inc
+++ b/recipes-extended/images/pelux-image-update.inc
@@ -19,7 +19,7 @@ IMAGE_DEPENDS = "${PELUXIMG}"
 SWUPDATE_IMAGES = "${PELUXIMG}"
 inherit swupdate
 
-FILESEXTRAPATHS_prepend := "${THISDIR}/pelux-image-update/:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/pelux-image-update/raspberrypi3:"
 SRC_URI = " file://sw-description "
 SRC_URI_append_raspberrypi3 = " file://emmcsetup.lua "
 


FlorentRevest:
That looks great to me, thanks @sashko !

However, I have never seen that "file:///" trick before. Out of curiosity, do you know why we need to add that extra slash?

sashko:
@FlorentRevest `file:///` will look for a file in the same directory; it can also be used for an absolute path, meantime `file://` will look for files in pathes set at `FILES` and `FILESEXTRAPATHS` variables.

But my fix was not entirely correct. I have updated the patch. Thanks.

FlorentRevest:
Well, I'm afraid that the updated patch you just merged wasn't correct either then... Or at least, it's not the expected behavior.

You basically forced every platform to use the sw-description of the raspberrypi3, while we actually need a different sw-description for each of them. It certainly hides the warning we had when trying to build for the nuc and arp but ultimately it will cause more problems than it solves right now.

Without your FILESEXTRAPATHS_prepend, bitbake was able to find and use the sw-description under the raspberrypi3/ directory when building for the rpi. The idea was that with one subdirectory per platform, bitbake would always use the right machine-specific sw-description.

The problem was: what happens in the "default case", when there is no sub-directory for the current MACHINE? As long as we don't actually build the recipe, it would simply cause a warning. I'm not saying it's an acceptable behavior, it was definitely a problem, but your patch doesn't seem quite right to me either.

sashko:
@FlorentRevest then yeah, my mess up.

GordanM:
@sashko so is there gonna be a fixup of this or should someone else do it? :)

sashko:
@GordanM, of course, I'll fix it.

PR #122, opened 2018-05-30
 Qt 5.11: Append rocko as a supported layer

thenor:


thenor:
Stopped the build since there are many builds in the queue and I can build manually first to test this out.

Gagi2k:
not sure why, but that didn't work for me, i needed to remove the LAYERSERIES_COMPAT_qt5-layer in meta-qt5/conf/layer.conf to make it work

thenor:
Yeah doesn't seem to have worked for me either.

thenor:
This worked locally at least.

conf/variant/common/local.conf
```diff
@@ -25,7 +25,8 @@ POKYQEMUDEPS_remove = "packagegroup-core-device-devel"
 NOHDD = "1"
 NOISO = "1"
 
-# Removing tar.bz2 image format as it is not used. Incase of need, it can be supplied through own local.conf
+# Removing tar.bz2 image format as it is not used.
+# If of needed it can be supplied through own local.conf
```
erikboto: Grammar got weird, maybe remove the "of".

conf/variant/common/bblayers.conf
```diff
@@ -1,3 +1,6 @@
+LAYERSERIES_CORENAMES_append = " rocko"
```
erikboto: Maybe a short comment on why this is done, since I guess it's quite unusual to change another layers config.

erikboto: (The comment above is meant for both lines here, not just the CORENAMES thing)

thenor: Oops I forgot to add my comment. I had it in another file, then moved it to this file.

conf/variant/common/local.conf
thenor: Yes. I will remove the "of".

PR #123, opened 2018-05-31
Revert "pelux-image-update: fix FILESEXTRAPATHS"

sashko:
This reverts commit cefa914dc715fd4d865a14000cef7c4633550427.


FlorentRevest:
I approve the PR. Just for further reference: this PR causes some warnings to be back when building for the NUC and ARP.
Those warnings would need a clever fix but unfortunately, off the top of my head, I can't think of a good solution to the problem you originally tried to solve. https://github.com/Pelagicore/meta-pelux/pull/120#issuecomment-393147779

PR #124, opened 2018-06-01
Use neptune3-ui as the new reference ui

Gagi2k:
This also adapts the bblayers-qtauto.conf to the new setup of meta-b2qt.
The neptune-ui.bbappend got renamed to neptune3-ui.bbappend and fixed.

### DIFF
--- a/conf/variant/common/bblayers-qtauto.conf
+++ b/conf/variant/common/bblayers-qtauto.conf
@@ -1,7 +1,8 @@
 
 BBLAYERS += "\
   ${BSPDIR}/sources/meta-qt5     \
-  ${BSPDIR}/sources/meta-boot2qt \
+  ${BSPDIR}/sources/meta-boot2qt/meta-boot2qt-distro \
+  ${BSPDIR}/sources/meta-boot2qt/meta-boot2qt \
 "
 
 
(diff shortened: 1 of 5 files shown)


conf/variant/common/bblayers-qtauto.conf
```diff
@@ -1,7 +1,8 @@
 
 BBLAYERS += "\
   ${BSPDIR}/sources/meta-qt5     \
-  ${BSPDIR}/sources/meta-boot2qt \
+  ${BSPDIR}/sources/meta-boot2qt/meta-boot2qt-distro \
```
GordanM: Cool stuff! I am wondering what is in this distro meta layer of boot2qt?

Gagi2k: it contains all the additional recipes. At the moment this is only needed because of the packagegroups. Either we stop using them and create our own or ask tqc to make them more generic and move them to the meta-boot2qt layer

GordanM: @Gagi2k thanks for the comment! Understood, maybe we can investigate this later on, it would be great to avoid any distro specific appendments from boot2qt so this kinda structuring is cool from their side. We can pick this up later!

Gagi2k:
yes, i think i need to rebase anyway.

Gagi2k:
I rebased, any way to start the test again ? I couldn't find a descriptive error there ?

Gagi2k:
Ahh and i don't think my change can be built easily as it needs a manifest update as well to point to the correct versions of meta-b2qt etc.

PR #127, opened 2018-06-11
WIP: SWUpdate support for intel-nuc

TSAnsari:


meta-intel-extras/recipes-support/grub/grub-efi/grubenv-copy.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
sashko: is service enabled?

meta-intel-extras/recipes-support/grub/grub-efi/grubenv-copy.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+if [ ! -d /boot/EFI/BOOT/ ]; then
```
sashko: isn't it excessive?

classes/core-image-pelux.bbclass
```diff
@@ -25,12 +25,13 @@ IMAGE_INSTALL += "\
     node-state-manager \
 "
 
-# OTA mechanism
-IMAGE_INSTALL_append_rpi += "\
+
+IMAGE_INSTALL += " \
     swupdate \
+    grub-efi \
```
thenor: Does this mean that grub would also be installed on rpi? Was that the intention?

meta-intel-extras/recipes-support/grub/grub-efi_%.bbappend
```diff
@@ -0,0 +1,27 @@
...
+
+SRC_URI += "\
+    file://grubenv \
+    file://grubenv-copy.sh \
+    file://grubenv-copy.service\
+"
+
+do_install_append() {
+    install -d ${D}${datadir}/
+    install -m 0755 ${WORKDIR}/grubenv ${D}${datadir}/
```
sashko: minor issue, but those slashes at the end are very unnecessary

classes/core-image-pelux.bbclass
TSAnsari: no, this was just to get things work quickly to test how it works with intel. This needs to be fixed.

meta-intel-extras/recipes-support/grub/grub-efi/grubenv-copy.sh
TSAnsari: This is needed because when the B partition is booted after swupdate, the boot partition is not mounted on it. I am not sure why but that is the reason to do this.

meta-intel-extras/recipes-support/grub/grub-efi/grubenv-copy.service
TSAnsari: Yes. The following line in grub-efi recipe makes sure that it is enabled

SYSTEMD_SERVICE_${PN} = "grubenv-copy.service"

sashko:
Obsolete.

PR #129, opened 2018-06-12
swupdate: add grubenv

sashko:
Added grubenv file with the default boot environment variable.


FlorentRevest:
Any reason why the grubenv has so many hashes?

sashko:
From the GRUB Manual:

> The environment block is a preallocated 1024-byte file

FlorentRevest:
Alright, it's probably something worth mentioning in a commit message

PR #130, opened 2018-06-12
LICENSE: add Luxoft Sweden copyright

sashko:
Signed-off-by: Oleksandr Kravchuk <dev@sashko.rv.ua>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
-Copyright (c) 2016-2017 Pelagicore
+Copyright (c) 2016-2017 Pelagicore AB
+Copyright (c) 2018 Luxoft Sweden AB
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


LICENSE
```diff
@@ -16,4 +17,4 @@ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
```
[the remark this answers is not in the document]
sashko: Hm... Must be my editor. Probably trailing line break.

LICENSE
```diff
@@ -1,4 +1,5 @@
-Copyright (c) 2016-2017 Pelagicore
+Copyright (c) 2016-2017 Pelagicore AB
```
thenor: Just FYI we decided not to use the range way of writing, so if you see this in any files you can remove it and leave only the first year.

sashko: Good to know. Do we update copyright with the current year or leave it as is?

PR #133, opened 2018-06-13
swupdate: install emmcsetup.lua on intel image

sashko:


recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup.lua
```diff
@@ -0,0 +1,85 @@
+function os.capture(cmd)
```
thenor: I'm assuming this is just copied from rpi, so nothing to review really?

sashko: Yes. Have you spotted any issues?

thenor: No, I just didn't feel a need to review the file in case it has already been reviewed for rpi. Also it would take me a little time to understand the code.

PR #134, opened 2018-06-13
layer.conf: add LAYERSERIES for sumo

sashko:

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -21,3 +21,5 @@ BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bbappend' % layer \
 # Do the same for recipes
 BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bb' % layer \
                for layer in BBFILE_COLLECTIONS.split())}"
+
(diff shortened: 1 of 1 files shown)


conf/layer.conf
```diff
@@ -21,3 +21,5 @@ BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bbappend' % layer \
 # Do the same for recipes
 BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bb' % layer \
                for layer in BBFILE_COLLECTIONS.split())}"
+
+LAYERSERIES_COMPAT_selftest = "sumo"
```
thenor: I see the PR title, but I still don't understand what this does. Could you explain briefly the purpose of this change? What does the selftest part mean?

sashko: It is a newish macro for `poky`; it basically states that this layer has been tested with `sumo` branch of poky.

This is not crazy necessary change, just a way to warn a user in case they're say compiling meta-pelux with `rocko` branch of `poky` and `morty` branch of `meta-openembedded`.

I have made this change mostly because one seems to need to make a change to create a branch on github.

thenor: Alright. Thanks for the explanation.

PR #136, opened 2018-06-15
grub-efi: add service that copies grubenv to boot

sashko:


FlorentRevest:
The "swupdate" tag before the commit name is a bit misleading as the commit doesn't touch the swupdate recipe but well, other than that it's ok.

PR #138, opened 2018-06-18
pelux.conf: fix DISTRO_NAME

sashko:


GordanM:
You can make two commits from this PR because you are changing two things. Otherwise, LGTM.

GordanM:
@sashko I've approved the pull request given that my comment was taken into account.

sashko:
@GordanM --

`You can...`

You need to be more precise in your language.

PR #142, opened 2018-06-21
swupdate: add comment regarding SRC_URI usage

TSAnsari:

### DIFF
--- a/recipes-extended/images/pelux-image-update.inc
+++ b/recipes-extended/images/pelux-image-update.inc
@@ -20,6 +20,14 @@ SWUPDATE_IMAGES = "${PELUXIMG}"
 inherit swupdate
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/pelux-image-update/:"
+
+# sw-description is a "hard requirement" (swupdate won't work without it,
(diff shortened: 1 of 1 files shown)


sashko:
it doesn't really do anything except unifying the style, right? could you change the title, please? the current one doesn't really make much sense.

TSAnsari:
Yes it just simplify/unify the code. Updated the commit message.

FlorentRevest:
I don't _fully_ agree with this PR because sw-description is a "hard requirement" (swupdate won't work without it, no matter what the platform is) whereas emmcsetup.lua could be kept out of the update artifacts for certain targets (for instance, if the disk is already partitioned or if we don't use an A+B partitioning scheme).

That being said, I don't really mind either... If you think it makes sense to have everything in one SRC_URI even if we loose some semantic with that, I'm fine with merging this PR.

TSAnsari:
@FlorentRevest Now that you explained why it was done like that before, it makes sense and I agree with your comments. So I want to keep it as it was. Only that the reason was not very obvious to me, so instead I will add a comment that explains the intention for someone thinking about it in the future.

TSAnsari:
closing this as consensus is that the code is self explanatory

Issue #143, opened 2018-06-23
Useless patch left in the layer


FlorentRevest:
I'm a bit confused by the existence of: layers/poky/patch/0001-Fix-automated-runtime-testing-using-SystemdbootTarge.patch
It looks like the patch doesn't apply on top of poky anymore and I have been able to run automated tests without using this patch. Does anyone know if it's necessary for anything? Should we re-work this patch or remove it?

GordanM:
Hey Florent, this patch was sent to upstream a while back by @erikboto, I believe if the patch doesn't apply anymore and the test are able to run on the target w/o it then this temp patch should be removed. Erik what are your thoughts about this?

Thanks!

PR #148, opened 2018-06-30
Remove an unused patch to Poky

FlorentRevest:
Remove an unused patch to Poky. Closes https://github.com/Pelagicore/meta-pelux/issues/143\#issuecomment-399759875


thenor:
I see that the build has failed, but have you tested building this manually, @FlorentRevest?

FlorentRevest:
@thenor I don't remember having tested it but the file wasn't used so I don't see any reason why it would fail. (yeah, we never know but...)

FlorentRevest:
I tried it out this week end and it works fine ;)

PR #150, opened 2018-07-04
intel: fix image install for grub-efi

thenor:
It seems the space before grub-efi was not included as it should.

### DIFF
--- a/conf/variant/intel/local.conf.sample
+++ b/conf/variant/intel/local.conf.sample
@@ -6,6 +6,4 @@ EFI_PROVIDER = "grub-efi"
 
 WKS_FILE = "mkefidisk-multiboot.wks"
 
-IMAGE_INSTALL_append = "\
-    grub-efi \
-"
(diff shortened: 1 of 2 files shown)


conf/variant/intel/local.conf.sample
```diff
@@ -6,6 +6,4 @@ EFI_PROVIDER = "grub-efi"
 
 WKS_FILE = "mkefidisk-multiboot.wks"
 
-IMAGE_INSTALL_append = "\
-    grub-efi \
```
TSAnsari: This is very strange as I can see several spaces before grub-efi. Anyway if it fixes the issue then I am fine with it.

thenor:
I didn't do a manual build yet, but I have seen that this was an issue before. I think the escape sign removes all whitespace, so I hope this is the issue.

thenor:
It seems that not everyone saw the same issue I did, even though I got it both in Jenkins and when building manually, but I still think it's good to have this change.

PR #155, opened 2018-07-06
Reverting cherry-picked commits from qt5.11 topic branch

fhajredini:
Changes reverted due to error on build.


fhajredini:
When i did the initial pull request i didnt test them cos i was quite confident it would work. The reason i want to revert changes is because the bug/issue is on the master. I will continue working on it so that i fix this.

fhajredini:
Apparently the error "internal compiler error killed (program cc1plus)" was related to my environment. The built had no failures on other environments therefore i'm closing this request.

PR #157, opened 2018-07-09
Add openssh server to qtauto image

TSAnsari:


sashko:
Could you please explain why this doesn't work?

Gagi2k:
Is there any reason why we just change the qtauto recipes ? Shouldn't the ssh-server be included already in the base image ?

TSAnsari:
@Gagi2k I don't see it added in base image. It was only added to the following dev images
core-image-pelux-qtauto-neptune-dev.bb
core-image-pelux-minimal-dev.bb

TSAnsari:
Based on the above discussion, that only dev image should contain ssh server,I am discarding this pull request

Gagi2k:
@TSAnsari sorry you missunderstood me, i meant why we are not adding it also to the base image... But i guess @FlorentRevest already answered that before...

PR #158, opened 2018-07-12
Sumo

sashko:


fhajredini:
I have already made a merge request to sumo as a qt5.11 bump and it contains 639ba61 commit. Maybe its better if we wait until we resolve the issue that we are having and then merge?

PR #159, opened 2018-07-17
Sumo

sashko:


Gagi2k:
In general i'm fine to have this in the sumo branch, but shouldn't you just merge or rebase the sumo branch on top of master instead ? (where the picked commits are already in ?)

sashko:
@Gagi2k yeah, that's probably better to do now

sashko:
@tsansari it is just a rebase on master.

sashko:
@fhajredini done

PR #160, opened 2018-07-20
grub-efi: fix package installation

sashko:


sashko:
@TSAnsari this issue shows up on sumo only.

PR #162, opened 2018-07-24
swupdate: Add support for arp target

TSAnsari:

### DIFF
--- a/recipes-extended/images/pelux-image-update.inc
+++ b/recipes-extended/images/pelux-image-update.inc
@@ -23,6 +23,7 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/pelux-image-update/:"
 SRC_URI = " file://sw-description "
 SRC_URI_append_raspberrypi3 = " file://emmcsetup.lua "
 SRC_URI_append_intel-corei7-64 = " file://emmcsetup.lua "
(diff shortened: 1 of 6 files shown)


sashko:
so, it's identical to the NUC implementation?

TSAnsari:
@sashko yes it is identical. Nothing new here.

recipes-extended/images/pelux-image-update.inc
```diff
@@ -23,6 +23,7 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/pelux-image-update/:"
 SRC_URI = " file://sw-description "
 SRC_URI_append_raspberrypi3 = " file://emmcsetup.lua "
 SRC_URI_append_intel-corei7-64 = " file://emmcsetup.lua "
+SRC_URI_append_arp = " file://emmcsetup.lua "
```
FlorentRevest: If the emmcsetup.lua and sw-description truely are the same (and will stay the same), I would have used the one from ``intel-corei7-64`` instead, for instance with

    SRC_URI_arp = "file://intel-corei7-64/sw-description file://intel-corei7-64/emmcsetup.lua"

But this is arguably an ugly line to write so if you think your solution is better, I'm also fine with it.

sashko: Yes, it's ugly :)

PR #163, opened 2018-07-24
Fix installation of core-image components

FlorentRevest:
While trying to debug a simple problem of "broken bluetooth" open for 2 months, I noticed that core-image-pelux lacked most of the basic features we could expect from it. For instance, a minimal bash was used as the init process, systemd wasn't used.
I tried to figure out what was going wrong and noticed that the poky's core-image's IMAGE_INSTALL wasn't included. It turns out that our usage of "IMAGE_INSTALL +=" overwrites [core-image's definition of "IMAGE_INSTALL ?=" ](http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/core-image.bbclass#n70)
Replacing those " +=" with "_append =" keeps the core packages such as systemd and bluetooth kernel modules and fixes many problems.

But a question remains... PELUX has been broken for months and no one noticed anything ? What the heck ?!

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -8,25 +8,25 @@ inherit core-image
 IMAGE_FEATURES += "package-management debug-tweaks"
 
 # Include softwarecontainer only if the process-containment feature has been enabled
-IMAGE_INSTALL += "\
+IMAGE_INSTALL_append = "\
     ${@bb.utils.contains("DISTRO_FEATURES", "process-containment", "softwarecontainer", "", d)} \
 "
 
 # Include bluetooth if the machine supports it (MACHINE_FEATURES), and it has
 # been selected in DISTRO_FEATURES.
-IMAGE_INSTALL += "\
+IMAGE_INSTALL_append = "\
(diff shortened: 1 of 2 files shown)


sashko:
> For instance, a minimal bash was used as the init process, systemd wasn't used.

Strange. This is not the case for the image I've built a few days ago.

FlorentRevest:
Were you using one of the -dev or Qt images? They probably pull some stuff in as dependencies of higher-level components. (for instance, the -dev image had systemd, probably as dependency of ssh, but it didn't have bluetooth drivers)

Also, here is a reference for my fix, that I forgot to mention: https://stackoverflow.com/a/46341457

sashko:
Yes, I found that as well. He didn't really answer the question. [Here is](https://lists.yoctoproject.org/pipermail/yocto/2012-June/007363.html) another one.

I am a bit confused. Let's double check everything tomorrow.

TSAnsari:
So, it means its a good idea to use _append instead of += unless one exactly know what one is doing?
I saw several uses of IMAGE_INSTALL += in meta-pelux layer. May be we should fix all of those.

TSAnsari:
I didn't see any issue in swupdate image as it uses a systemd service to copy some grub env files. Perhaps it works since the recipe added for grub-efi inherits systemd.

PR #167, opened 2018-08-06
linux-intel: fix SRCREV

sashko:
meta-intel currently points to nonexistent commit in meta-intel.
Temporarily fix by hardcoding SRCREV to latest stable Linux kernel
version of the 4.9 branch.


TSAnsari:
I am running the build with your patch and it seems to be running fine. 
We should create a Jira ticket for when the issue is fixed in meta-intel branch so that we don't forget to revert this patch in our layer.

thenor:
I agree with @TSAnsari that it's good to create a user story as soon as possible so we don't forget that this is hardcoded.

PR #168, opened 2018-08-07
Add CAN utilities and Socket CAN modules

TSAnsari:

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -1,5 +1,6 @@
 #
 #   Copyright (C) 2017 Pelagicore AB
+#   Copyright (C) 2018 Luxoft Sweden AB
 #   SPDX-License-Identifier: MIT
 #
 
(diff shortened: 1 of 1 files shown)


fhajredini:
`# Copyright (C) 2017 Pelagicore AB`
Maybe this is something we can update as well.

sashko:
lower case, Tariq :)

classes/core-image-pelux.bbclass
```diff
@@ -1,5 +1,5 @@
 #
-#   Copyright (C) 2017 Pelagicore AB
+#   Copyright (C) 2018 Luxoft Sweden AB
```
sashko: You cannot change copyright!

thenor: When doing this a few weeks ago the consensus is that if the file has "Pelagicore" then we add a line with "Luxoft Sweden". If there is a line with a previous year and "Luxoft Sweden" then leave as is, since the Copyright is valid for many many years.

sashko: Correct. But this is not what has been done.

thenor: Yes, that was also my point.

TSAnsari: So in this case I should revert back to "Copyright (C) 2017 Pelagicore AB"? Its not clear to me from the above comments what to do.

thenor: Sorry about the confusion, no I mean that you should add a line with "Luxoft Sweden" and 2018.

PR #178, opened 2018-08-28
gammaray: fix rpath

sashko:
Fixes the following issue:

do_package_qa: QA Issue: package gammaray contains bad RPATH
in file .../gammaray/usr/lib/libgammaray_core-qt5_10-x86_64.so.2.9.0


kanavin:
Please note that this is only a workaround for the broken build system - these paths shouldn't appear in binaries in the first place.

sashko:
@kanavin completely agree - this should be fixed in the project's build system, and it seems to be in the newer versions.

PR #180, opened 2018-08-31
swupdate: disable systemd service

sashko:
Since we don't need swupdate service to start on system boot, disable it.


FlorentRevest:
Yeah we don't use that service yet. The idea is to have swupdate running in "surricata" mode (in background) and checking for updates regularly but then it would need some changes to the script actually ran.
As is, this service is not useful.

PR #184, opened 2018-09-07
Build sumo image with security flags

sashko:


kanavin:
Looks right to me. You might want to raise the need to except with softwarecontainer people.

sashko:
Turns out, the latest softwarecontainer revision builds perfectly fine. I'll bump the recipe and update this PR.

PR #189, opened 2018-09-13
boot2qt/connman: hardcode version of the patch

fhajredini:
boot2qt has updated the recipes to look for the latest connman patch(1.35). This patch hasn't been added to the files yet. Until that is done, the patch (1.34) should remain hardcoded.  


recipes-temporary-patches/boot2qt/connman_%.bbappend
```diff
@@ -0,0 +1,5 @@
+
+FILESEXTRAPATHS_prepend := "${BSPDIR}/sources/meta-boot2qt/meta-boot2qt-distro/recipes-connectivity/connman/files:"
```
TSAnsari: This doesn't seem right. The "sources" directory may or may not exist, so hard-coding it here doesn't seem to be a good idea.

TSAnsari: Though I am not sure since I see such usage in many places in meta-pelux as well as other layers

TSAnsari: Does it work, if we do something like this?
${LAYERDIR}/../meta-boot2qt/meta-boot2qt-distro/recipes-connectivity/connman/files:

recipes-temporary-patches/boot2qt/connman_%.bbappend
```diff
@@ -0,0 +1,5 @@
...
+
+SRC_URI += "file://0001-disable-connman-when-booting-over-nfs-1.34.patch"
```
TSAnsari: I assume using += will just append this file and the 1.35.patch will still be there, so it will try to find it. Have you tested this that it works like this?

recipes-temporary-patches/boot2qt/connman_%.bbappend
sashko: just copy the patch to the same dir with bbappend

recipes-temporary-patches/boot2qt/connman_%.bbappend
sashko: Yeah, use `_append`

TSAnsari: I meant to say that SRC_URI will contain both 
1.35 and 1.34 patch and might fail again when trying to look for 1.35. So just test locally that it actually works

fhajredini: Yeah, that was the case. It was still looking for 1.35

fhajredini:
@sashko @TSAnsari @martin-ejdestig Please review this issue.

sashko:
it's more like add patch for 1.35 :)

PR #190, opened 2018-09-14
gammaray: add bbappend with temp rpath fix

sashko:

### DIFF
--- a/recipes-temporary-patches/gammaray/gammaray_git.bbappend
+++ b/recipes-temporary-patches/gammaray/gammaray_git.bbappend
@@ -1,0 +1,9 @@
+do_install_append_intel-corei7-64() {
+    chrpath -d ${D}/usr/lib/libgammaray_core-qt5_11-x86_64.so.2.9.1
+}
+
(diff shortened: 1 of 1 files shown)


sashko:
to clarify the change: rpath issue exists on the latest version of `gammaray`; `meta-boot2qt` will have to be bumped to use `Qt 5.11` on `sumo`.

kanavin:
The error is most likely a package_qa failure, which has a check specifically for this:
https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/insane.bbclass#n98

I have to agree: there is an issue in the way the component is built, and fixing it after the fact at installation time is not right.

sashko:
@kanavin I understand that bbappends with those kinds of fixes should be temporary solutions until the proper solution is mainstreamed, but I don't always have time (or just not in a mood) to fix it in the mainstream.

Can you suggest us a better way of working around those kinds of issues? Thanks.

kanavin:
It's fine as a workaround if you don't have time, as long as it contains a link to the upstream ticket to fix this properly :)

PR #191, opened 2018-09-14
qtapplicationmanager: fix path to library

sashko:
Fixed path to the library and cleaned up syntax a bit.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/qtapplicationmanager-sc.inc
+++ b/layers/b2qt/recipes-qt/automotive/qtapplicationmanager-sc.inc
@@ -1,5 +1,6 @@
 #
 #   Copyright (C) 2017 Pelagicore AB
+#   Copyright (C) 2018 Luxoft Sweden AB
 #   SPDX-License-Identifier: MIT
 #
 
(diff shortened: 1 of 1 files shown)


sashko:
@martin-ejdestig the path has changed on `Qt 5.11`. This change is needed to bump `Qt` to `5.11` on `sumo`.

martin-ejdestig:
I would split this commit in two. One commit for the actual fix and one with the cleanups.

We are not hit by this now because we are not on 5.11 yet? Or we do not build qtapplicationmanager?

sashko:
@mardy we're not hit by this since we're not using 5.11 yet (we're on 5.10 atm)

martin-ejdestig:
So if we commit this it will be broken until we update?

sashko:
@martin-ejdestig well, if you manually fetch latest meta-pelux then yes, but since manifest will not be updated, CI/CD builds will work.

martin-ejdestig:
But your plan is to update immediately after, or? If not, are we OK with not being able to update meta-pelux until that is done?

sashko:
@martin-ejdestig the plan is to bump `meta-boot2qt` and `meta-qt5` right after all the necessary fixes are delivered to `meta-pelux`.

PR #194, opened 2018-09-17
Mopidy integration

FlorentRevest:
This PR integrates the Mopidy media playback service in PELUX. It also integrates PulseAudio and enables sound on the RaspberryPi.

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -32,6 +32,11 @@ IMAGE_INSTALL_append = "\
     can-utils \
 "
 
+# Media playback daemon
+IMAGE_INSTALL_append = "\
+    mopidy \
+"
+
 # OTA mechanism
 IMAGE_INSTALL_append_rpi = "\
     swupdate \
--- a/meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
+++ b/meta-rpi-extras/recipes-bsp/boot-files/rpi-config_%.bbappend
@@ -12,3 +12,8 @@ GPU_MEM="${@bb.utils.contains("BBFILE_COLLECTIONS", "b2qt", 512, 128, d)}"
 # commandline.txt) enables serial console from the kernel.
 #
 ENABLE_UART="1"
+
+do_deploy_append_raspberrypi3() {
+    echo "# Enable audio (loads snd_bcm2835)" >> ${DEPLOYDIR}/bcm2835-bootfiles/config.txt
+    echo "dtparam=audio=on" >> ${DEPLOYDIR}/bcm2835-bootfiles/config.txt
+}
--- a/recipes-devtools/python/python-pykka_1.2.1.bb
+++ b/recipes-devtools/python/python-pykka_1.2.1.bb
@@ -1,0 +1,10 @@
+SUMMARY = "Pykka is a Python implementation of the actor model"
+HOMEPAGE = "https://www.pykka.org/"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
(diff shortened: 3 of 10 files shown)


kanavin:
Why add gstreamer-0.10 stuff? Does something depend on it? This version of gstreamer has been out of support for a very long time.

kanavin:
Also I would recommend that you do not add new python 2.x dependencies. Python 2.x will be unsupported (and possibly removed from oe-core) soon.

FlorentRevest:
> Why add gstreamer-0.10 stuff? Does something depend on it? This version of gstreamer has been out of support for a very long time.

Very valid point! Sorry for that.

I issued a new build with sumo and this bbappend was not necessary anymore. It was probably something to do with my rocko build or a leftover from an early experiment I must have had done. I'm quite sure I didn't write that bbappend by pure pleasure but it was a mistake that shouldn't have stayed in the PR, indeed...

> python-pykka: add recipe

I assumed you wanted me to reword my commit messages with that format. I also changed the mopidy commit if that's what you wanted.

> Also I would recommend that you do not add new python 2.x dependencies. Python 2.x will be unsupported (and possibly removed from oe-core) soon.

That's also a valid point and I really share your concerns, but:
- The media playback service has been chosen a long while ago in an Architecture Meeting, before I did work for this company. From what I could read, Mopidy was by far the solution that would fit PELUX the best. Anyway, the choice of this media playback service wasn't mine and I just integrated what was on the roadmap.
- A migration to Python 3 has been in their pipeline since 2014: https://github.com/mopidy/mopidy/issues/779 From what I could gather, it is almost complete but still a work in progress. I really hope that they will have pressure from their dependencies (such as pylast) and be required to move to Python 3.
- When you say oe-core could drop Python 2 "soon". Do you have an idea of what you mean with "soon" ? Any distribution that decides to drop Python 2 knows it causes hell on earth. I'm sure there will be a meta-python2 "answer" to this. Trust me, I'd much rather avoid using it in PELUX but my point is: there will always be a workaround.

kanavin:
I'm glad mopidy is moving to 3.x. I should clarify 'soon': all upstream for python 2.x support, including security fixes will cease on 1st of january 2020. I'm sure you'll be able to get a python 2.x recipe one way or another, but if it proves to be too much of a pain to keep in oe-core (because for instance it starts regressing with new toolchain or glibc versions or other dependencies) it will be downgraded to some 'community' layer with no promises or guarantees. And generally taking recipes like that into your distro is not a good idea in the long term.

FlorentRevest:
Hopefuly, by the time upstream Python stops maintaining 2.x, Mopidy will be done with the few remaining tasks left to port it to 3.x.
The project is really quite active and I'm confident they will handle this issue before problems start arising in PELUX.
In the meantime, we can keep an eye on https://github.com/mopidy/mopidy/issues/779 and cross fingers... :) 
Does this sound right to you?

sashko:
@FlorentRevest be aware that I modified some commit messages (yes, I know) and forced pushed to your `mopidy` branch.

How can we try it out? I would like to play a Spotify song.

sashko:
@FlorentRevest cherry-pick those patches to `sumo` when you have a chance, please.

FlorentRevest:
@sashko okay, done!

hugovk:
> A migration to Python 3 has been in their pipeline since 2014: mopidy/mopidy#779 From what I could gather, it is almost complete but still a work in progress. I really hope that they will have pressure from their dependencies (such as pylast) and be required to move to Python 3.

Python 2.7 (and therefore all Python 2) reaches EOL on January 1st, 2020. 

Pylast will drop support for Python 2.7 before that, perhaps this year.

https://github.com/pylast/pylast/issues/265

FlorentRevest:
Thanks for the clarifications @hugovk! Hopefully, Mopidy will follow soon after.

PR #202, opened 2018-09-27
neptune3-ui: checkout assets from git lfs

sashko:
Neptune 3 UI assets are not stored at Git LFS, and current recipe from
meta-boot2qt does not seem to check them out at any stage. This change
fixes it.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend
+++ b/layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend
@@ -1,5 +1,6 @@
 #
 #   Copyright (C) 2017 Pelagicore AB
+#   Coprright (C) 2018 Luxoft Sweden AB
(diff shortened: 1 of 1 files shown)


Gagi2k:
Mhh, we had the same step in the beginning as well but tqc manages it to somehow get the assets without doing this check, so i guess it would be better to sync with TQC, why the recipe works for them without this, especially as it also worked for PELUX until recently ?

sashko:
@Gagi2k do you mind pointing me into the file which has contained this step? Perhaps I could go through git history and see what it was replaced with...

I only found this in `meta-boot2qt/conf/layer.conf`:

> HOSTTOOLS += "git-lfs"

Gagi2k:
neptune3 used git-lfs since the beginning. AFAIK the step has never been part of the recipe, but it still worked for them for some reason...

sashko:
Then my guess is that we build it in different ways.

PR #210, opened 2018-10-04
local.conf.sample: set PREFERRED_PROVIDER for iasl-native

sashko:
Bitbake produces the following warning:

  NOTE: Multiple providers are available for iasl-native (iasl-native, acpica-native)
  Consider defining a PREFERRED_PROVIDER entry to match iasl-native


thenor:
I would prefer that you remove the "Bitbake gives the following warning" part of the commit message. I guess this is to solve some compatibility issue with an updated layer or something.

PR #213, opened 2018-10-24
Fix container & appman config files

TSAnsari:
This fix enable the Neptune3 calendar application
to run inside a Software Container. Also updated
the gateway config and environment variable for
calendar application inside the container to
access the session dbus on host to communicate
with the neptune3 ui on the host.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/qtapplicationmanager/sc-config.yaml
+++ b/layers/b2qt/recipes-qt/automotive/qtapplicationmanager/sc-config.yaml
@@ -6,7 +6,7 @@ formatType: am-configuration
 ---
 containers:
   selection:
-    - com.pelagicore.radio: "softwarecontainer"
+    - com.pelagicore.calendar: "softwarecontainer"
     - "*": "process"
 
   softwarecontainer:
--- a/recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
+++ b/recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
@@ -8,6 +8,14 @@
           "id": "dbus",
           "config": [
             {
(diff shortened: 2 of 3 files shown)


layers/b2qt/recipes-qt/automotive/qtapplicationmanager/sc-config.yaml
```diff
@@ -6,7 +6,7 @@ formatType: am-configuration
 ---
 containers:
   selection:
-    - com.pelagicore.radio: "softwarecontainer"
```
sashko: so what happens to the radio app?

TSAnsari: There was no radio app in neptune3. There is webradio app but it doesn't matter what app to run in container as long as the configuration is correctly setup for that app

sashko: explain in the title what you fix; fix, do, change, make are not good examples of words to use

sashko: moreover, changes can be split

thenor: Is this just some example config? Is that why it doesn't matter if it's changed?

TSAnsari: This is not an example config, but one can decide to run any application inside container. So we can have more applications with "softwarecontainer" value set in this configuration. For PELUX we are just running one of the several applications inside the container. The important thing is that whatever application we decide to run inside a container, the access to resources needed by that application must be set in the service manifest file.

PR #223, opened 2018-11-08
conf-notes: add variant name to text

thenor:
To make it clear which variant will be built, print it out when the
poky build-env is being sourced.

### DIFF
--- a/conf/variant/arp-qtauto/conf-notes.txt
+++ b/conf/variant/arp-qtauto/conf-notes.txt
@@ -1,4 +1,4 @@
-Available PELUX images:
+Available PELUX images for arp-qtauto:
 * core-image-pelux-minimal
 * core-image-pelux-minimal-dev
 * core-image-pelux-qtauto-neptune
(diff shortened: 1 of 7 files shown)


fhajredini:
I think its actually easier with rpi-qtauto and just as much understandable. But yet, either way works.

rpannek:
Can't you just put the variable name of the variant there?

thenor:
@rpannek I don't understand if you mean the full name, as @sashko mentioned, or the variant as @fhajredini mentioned. The text I added was the name of the variant used for the directory.

jeena:
I meant whatever you get from the VARIANT variable. But it doesn't really matter.

rpannek:
I meant whatever you get from the VARIANT variable. But it doesn't really matter.

thenor:
@rpannek Yes, that's what I'm using now. It would have been nicer to make it a script, but at the moment it's just a txt, so I just wanted to add this quickly.
I'm just going to pull this in now. We can change it if it's not enough information.

PR #226, opened 2018-11-13
Bump sumo

thenor:


thenor:
@TSAnsari Let me know if I also need any softwarecontainer changes as well.

thenor:
This builds locally with the sumo base. I know we have issues in Jenkins which results in internal layer bumps being built on a master base regardless of the PR base.

PR #237, opened 2018-12-04
Revert "qtbase: fix intermittent rpi-qtauto gammaray build failures"

sashko:
This issue has been fixed in the mainline meta-qt5@3820e2d1b2a8f961980.

This reverts commit dce88dd84a2efc38a60d48a48e6a22752591b8ff.


martin-ejdestig:
I have not looked at the upstream commit but might be a good idea to verify that the bitbake commands I wrote in the reverted commit's commit message does not give you any error.

martin-ejdestig:
Looking at the console output of the failed build, it looks like the same error is reintroduced. Do not know why. We were lucky that it just happened to fail.

Anyway, please make sure that the steps to reproduce are not failing before merging this (do not rely in Jenkins failing, because it's random).

PR #238, opened 2019-01-18
qtapplicationmanager: Removed explicit hash

JEderonn:
This fix is required to make appman and neptune 3 ui be in sync.

Since the manifest has been updated to 5.12 this patch is not
needed anymore.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/qtapplicationmanager_git.bbappend
+++ b/layers/b2qt/recipes-qt/automotive/qtapplicationmanager_git.bbappend
@@ -6,6 +6,4 @@
 containment = "${@bb.utils.contains('DISTRO_FEATURES', 'process-containment', 'sc', 'noop', d)}"
 require qtapplicationmanager-${containment}.inc
 
-SRCREV = "5c08826b6c794fe67ed2ebecca4a53ecf918ecca"
-
 ALLOW_EMPTY_${PN}-tools = "1"


Gagi2k:
This is the change which added the specific appman version: https://github.com/Pelagicore/meta-pelux/commit/bbe03e789b7b9678cd905a515e8cd21fee93ff92

As you might still want to use this fix, maybe create a patch which just adds the changes you need ? Or disable the container-plugin config ?

Gagi2k:
Or update to the 5.12 release version of appman straight away

JEderonn:
@Gagi2k : The hash you mention points to http://code.qt.io/cgit/qt/qtapplicationmanager.git/commit/?h=dev&id=5c08826b6c794fe67ed2ebecca4a53ecf918ecca "Make tst_main insignificant for the Luxoft CI" which is the commit after "SoftwareContainer fixes from the PELUX team". So if we update the recipe to 5.12, these commits will be included.

The current version of **pelux-manifest** uses for meta-boot2qt hash http://code.qt.io/cgit/yocto/meta-boot2qt.git/commit/?h=sumo&id=880ecb7a6e993a986d24c9a8b813bf2a48f51d24. In this layer, **qtapplicationmanager** uses hash 941d493f47d1e15430ed062ea0aa4b1f7c087a04 (dev-branch) "Fix blank line output in controller's list-* commands" and **neptune 3 ui** uses hash d8332c73ccda36e1da2661317d7ed3c787b17c68 (dev-branch) "[systemui] remove the window manager connection from apppopupsstore" which is VERY old, from 2018-09-25.

In short term, I would propose to make the following changes:
1. Merge this fix
2. Update the pelux-manifest so meta-boot2qt is in sync with QtAS 5.12

Gagi2k:
For master (which i guess will be PELUX 4.0 ?) this is ok, but what about PELUX 3.0 ? If this will remain using this sha versions you will always end up with problems...

In my opinion PELUX 3.0 should either be updated to use QtAS 5.11 or be updated to QtAS 5.12 (which doesn't need Qt 5.12, but would be handy)

PR #239, opened 2019-01-18
Jenkinsfile: Change variant list handling

fhajredini:


fhajredini:
This fixes the isse on the beginning when the build doesnt even start. Will merge it so we can reproduce any other error

PR #243, opened 2019-01-23
neptune.service: update to use correct config filename

TSAnsari:


martin-ejdestig:
Does this not have to be part of a commit/branch where you update Neptune as well to not cayse intermittent breakage? Or when did it change?

martin-ejdestig:
Ah, ok it should go together with a manifest bump in pelux-manifests.

TSAnsari:
@martin-ejdestig yes correct

PR #244, opened 2019-01-28
neptune.service: workaround for neptune3 ui

TSAnsari:
With neptune3, multiple applications in neptune ui
runs as separate processes. On rpi this causes some
performance issues since it doesn't have a very good
GPU. This causes a very low frame rate as well as
issues with correct rendering of the applications.
A workaround is added which runs one application at
a time as a single process. This fixes the performance
and rendering issues. This is a temporary fix, until
a solution is provided in neptune for this issue.


sashko:
that's one ambiguous title :/

TSAnsari:
Updated to better commit message and description

Gagi2k:
> A workaround is added which runs one application at a time as a single process

That's not really true. ALL applications incl. the system-ui are run in one-process. It's like loading the QML code directly in the system-ui with a Loader element

TSAnsari:
@Gagi2k yes I saw that and fixed it in the new message. Thanks

PR #245, opened 2019-01-29
core-image-pelux: temporarily remove mopidy

sashko:
Mopidy installation is broken in thud due to python2 support being
dropped by Yocto and mopidy not suppororting python3 just yet.

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -33,9 +33,9 @@ IMAGE_INSTALL_append = "\
 "
 
 # Media playback daemon
-IMAGE_INSTALL_append = "\
-    mopidy \
-"
+#IMAGE_INSTALL_append = "\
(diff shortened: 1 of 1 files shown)


classes/core-image-pelux.bbclass
```diff
@@ -33,9 +33,9 @@ IMAGE_INSTALL_append = "\
 "
 
 # Media playback daemon
-IMAGE_INSTALL_append = "\
-    mopidy \
-"
+#IMAGE_INSTALL_append = "\
+#    mopidy \
+#"
```
thenor: In my opinion this should be removed, not commented out, and then the commit can be reverted once Mopidy is supported again.

thenor:
I guess the removal of Mopidy would only be done on master if the bump to thud is done. So should this PR be closed for now?

PR #247, opened 2019-02-07
local.conf.samples: whitelist commercial for faad2

sashko:
faad2 is licensed under the GPLv2 and a commercial license for a
commercial use.

Since commercial is set in the recipe and we are not using it in the
commercial purposes, we temporarily whitelist license for faad2 package
until a better solution is found.

### DIFF
--- a/conf/variant/rpi/local.conf.sample
+++ b/conf/variant/rpi/local.conf.sample
@@ -1,6 +1,7 @@
-
 require conf/variant/common/local.conf
 
 MACHINE = "raspberrypi3"
 
 RPI_USE_U_BOOT = "1"
+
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
Why is this only required for rpi? Disabled for Intel? Disable for rpi as well?

(Also, commercial use should be OK as well as long as one abides by the rules of the GPL. We ship plenty of other packages that are GPL only so not a problem in itself. That is, you commit message may be a bit misleading?)

martin-ejdestig:
And hum, is the license just wrong in the recipe? Just because it is available under a commercial license does not mean that is something that has to be set there?

PR #251, opened 2019-02-15
local.conf.sample: auto-load kernel modules

sashko:
Auto load kernel modules necessary for proper functioning of the
Neptune 3 UI.


sashko:
> Were these modules missing before? or is it something new which is now required on thud?

No, they are simply not loaded. I will try to figure out why, but right now would like to have them loaded, so Neptune starts and works.

PR #253, opened 2019-02-19
grub-efi: add fstab to meta-intel to mount EFI partition 

TSAnsari:
A check prevented mount of EFI partition to /boot directory,
incase /boot directory already existed. This eventually result
in failing to copy the grubenv file to EFI partition. This
check is removed to fix this issue.

### DIFF
--- a/meta-intel-extras/recipes-core/base-files/base-files/intel-corei7-64/fstab
+++ b/meta-intel-extras/recipes-core/base-files/base-files/intel-corei7-64/fstab
@@ -1,0 +1,3 @@
+/dev/root            /                    auto       defaults              1  1
+/dev/sda1	/boot	vfat	defaults	0	0
+/dev/sda4	swap	swap	defaults	0	0
--- a/meta-intel-extras/recipes-core/base-files/base-files_%.bbappend
+++ b/meta-intel-extras/recipes-core/base-files/base-files_%.bbappend
@@ -1,0 +1,8 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
+#   SPDX-License-Identifier: MIT
+#
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+SRC_URI += "file://fstab"
+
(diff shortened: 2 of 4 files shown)


martin-ejdestig:
I do not understand. "if [ ! -d "... checks if the directory does not exist, and if so, mounts.

TSAnsari:
> I do not understand. "if [ ! -d "... checks if the directory does not exist, and if so, mounts.

That was the bug :)

martin-ejdestig:
Before it was "if directory does not exist, mount so that it does". You change it to "mount regardless of whether it exists or not". So, why does it help to remove the check?

martin-ejdestig:
You also add a mkdir. Is that perhaps the real bug?

(Also, I am not sure in what context this script executes. Why is there no /boot?)

TSAnsari:
> You also add a mkdir. Is that perhaps the real bug?
> 
> (Also, I am not sure in what context this script executes. Why is there no /boot?)

Perhaps there is no need to "mkdir -p /boot". We can assume it will exist.

martin-ejdestig:
Then... your patch does nothing... ?

TSAnsari:
> Then... your patch does nothing... ?

So  the issue was that /boot/EFI/BOOT was created by grub-efi recipe in poky. So it was not the mounted from EFI partition. So the -d check assumed that the EFI partition is already mounted and didn't mount the EFI partition.

martin-ejdestig:
There we go! (Write something like that in the commit message please! :)

martin-ejdestig:
Or hum... is the bug that the directory is created elsewhere?

sashko:
what would happen if we've created the directory if it doesn't exist?

martin-ejdestig:
Tariq explained the issue this script is trying to solve to me more in detail.

I think the problem is that the service it is run from (see grubenv-copy.service in same directory as grubenv-copy.sh ) does not have "After=local-fs.target" in it so /boot, that we have in /etc/fstab, may not be mounted when the script runs.

(The real fix, of course, is to not have the script and service at all and just make sure grubenv is put on the /boot partition instead of the rootfs when creating the image. I do not know the details of how that works though.)

sashko:
Wouldn't this break `swupdae` since `sda1` is not always the `/` partition?

TSAnsari:
It isn't / partition. sda1 is always the /boot partition. Or may be I didn't understand your question? Thanks


BR

Tariq

________________________________

martin-ejdestig:
Sorry, did not understand that you were referring to my "issue". The question was if we should really add a manual copy of /etc/fstab like this when it is generated when we build already. Would be **really** nice not to...

PR #254, opened 2019-02-21
systemd: enable myhostname packageconfig 

sashko:
bootchart2 recommends to set CONFIG_TASKSTATS to 'y'.

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -1,4 +1,3 @@
-
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 # This sets up a default ethernet on DHCP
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
Which incarnation of bootchart are you referring to? Can not see any mention in https://github.com/systemd/systemd-bootchart . Can you give a reference please? (Would be nice to have in commit message.)

sashko:
> Which incarnation of bootchart are you referring to?

https://github.com/dpaleino/bootchart2/

martin-ejdestig:
Could you perhaps make this clear in the commit message by e.g. linking to the README of bootchart2 ("See https://github.com/...")?

martin-ejdestig:
Btw, do we really need this (bootchart2)? systemd has systemd-analyze which can be used to plot a graph.

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -20,5 +19,7 @@ do_install_append() {
     done
 }
 
+EXTRA_OECONF = "myhostname"
```
martin-ejdestig: It looks to me like this means "-Dmyhostname=true" is passed to Meson when systemd is built, which means "nss-myhostname support".

What you really want, I think is for "-Dhostnamed=true" ("install the systemd-hostnamed stack") to be passed, which means this should read:

EXTRA_OECONF = "hostnamed"

Should it not also be an append (+=)?

And hum... should it not be appended to PACKAGECONFIG? (Is EXTRA_OECONF even relevant here? Only for configure and should be explicit configure flag? Since systemd is built with Meson it should be EXTRA_OEMESON if anything?)

PR #256, opened 2019-02-26
python-tonado: add recipe

sashko:
python-tornado 4.x is needed by Mopidy. meta-oe is currently provides
version 5 only.

### DIFF
--- a/recipes-python/tornado/python-tornado45.bb
+++ b/recipes-python/tornado/python-tornado45.bb
@@ -1,0 +1,25 @@
+#
+#   Copyright (C) 2018 Koen Kooi
+#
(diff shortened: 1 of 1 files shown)


TSAnsari:
Is this work  to get python2 on thud for Mopidy?

sashko:
> Is this work to get python2 on thud for Mopidy?

Not the python2, but the specific package, `python-tornado` that `mopidy` depends on.

TSAnsari:
It would be good to give some more background details for this change in the commit message. For example that yocto has removed python2 in thud and mopidy doesn't support python3 and that this should be removed once Mopidy supports python3 yada yada.

PR #257, opened 2019-02-26
mopidy: add dependancy on python-tornado45

sashko:
Make sure that python-tornado v4.5 provided by meta-pelux will be
on the image instead of newer versions provided by meta-openembedded.


martin-ejdestig:
Why?

martin-ejdestig:
(Current me knows why, but me a year from now may not. Neither may 1 year later Oleksandr.)

TSAnsari:
It would be good to add to commit message why we needed an older version

PR #259, opened 2019-02-27
swupdate: run as a service

TSAnsari:

### DIFF
--- a/recipes-extended/images/pelux-image-update/arp/emmcsetup1.lua
+++ b/recipes-extended/images/pelux-image-update/arp/emmcsetup1.lua
@@ -1,0 +1,130 @@
+function os.capture(cmd)
+	local f = assert(io.popen(cmd, 'r'))
+	local s = assert(f:read('*a'))
+	f:close()
+	return s
+end
+
+function file_exists(name)
+	local f=io.open(name,"r")
+	if f~=nil then io.close(f) return true else return false end
+end
+
+function cmdexec(cmd)
+	local ret, s, status = os.execute(cmd)
+	if (status ~= 0) then
+		return false, cmd .. " return with error " .. status
+	end
+
+	return true,""
+end
+
+function preinst()
+	local out
+	local s1
+	local ret
+
+	local log = os.tmpname()
+
+	local eMMC = "/dev/sda"
+	ret = file_exists("/dev/sda")
+
+	if (ret == false) then
+		return false, "Cannot find eMMC"
+	end
+
+	ret, out = cmdexec("/usr/sbin/sfdisk -d " .. eMMC .. "> /tmp/dumppartitions")
+	if (false == ret) then
+		return ret, out
+	end
+
+	-- check if there are two identical partitions
+	-- and create the second one if no available
+	f = io.input("/tmp/dumppartitions")
+	fo = io.output("/tmp/partitions")
+	t = f:read()
+	found = false
+	while (t ~= nil) do
+		j=0
+		j=string.find(t, "/dev/sda3")
+		ret, out = fo:write(t .. "\n")
+		if (ret == nil) then
+			fo:close()
+			f:close()
+			return false, out
+		end
+		if (j == 1) then
+			found=true
+			break
+		end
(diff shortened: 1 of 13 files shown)


sashko:
Did you write or copied those lua scripts from somewhere?

TSAnsari:
The scripts are from before. They had preinst function implemented. I have only added postinst function to them. Also split the single emmcsetup.lua scripts into 2 emmcsetup1.lua and emmcsetup2.lua as logic in postinst changes for different partitions. 

And there was no way of creating a helper lua script that contained common functionality, hence, this duplication of code.

recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup1.lua
```diff
@@ -0,0 +1,115 @@
...
+
+function file_exists(name)
+	local f=io.open(name,"r")
+	if f~=nil then io.close(f) return true else return false end
+end
+
+function cmdexec(cmd)
+	local ret, s, status = os.execute(cmd)
+	if (status ~= 0) then
+		return false, cmd .. " return with error"
```
JEderonn: would be good to indicate the value of status in the error message

recipes-extended/images/pelux-image-update/arp/emmcsetup1.lua
```diff
@@ -81,5 +81,35 @@ end
 function postinst()
 	local out = "Post installed script called"
 
+	ret, out = cmdexec("mkdir -p /tmp/mountedsda2")
+	if (ret == false) then
```
JEderonn: A general comment. It's good practice to do comparisons such as
IF (CONSTANT == VARIABLE)
i.e.
if (false == ret)
if (1 == j)
etc

Yes, good compilers will find and notify about potential errors like:
if (ret = false)
but in some cases the code will actually be easier/faster to read if the variable is put at the beginning.

recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup1.lua
```diff
@@ -0,0 +1,115 @@
...
+	if (found) then
+		f:close()
+		fo:close()
+		return true, out
+	end
+
+	start=start+size
+	partitions = eMMC .. "3 : start=    " .. string.format("%d", start) .. ", size=  " .. size .. ", type=83\n"
+
+	fo:write(partitions)
```
JEderonn: what if an error occur when writing partitions?

recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup1.lua
```diff
@@ -0,0 +1,115 @@
...
+	fo:close()
+	f:close()
+
+	out = os.capture("/usr/sbin/sfdisk --force " .. eMMC .. " < /tmp/partitions")
```
JEderonn: handle any error

recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup1.lua
```diff
@@ -0,0 +1,115 @@
...
+
+	-- use partprobe to inform the kernel of the new partitions
+
+	cmdexec("/usr/sbin/partprobe " .. eMMC)
```
JEderonn: what if the command fails? handle error

TSAnsari:
> There are a lot of duplicate code (lua-code). Would it be possible to refactor the code/files, and put the duplicate code into a file with common functions taking the variants as arguments?

Yes I tried to find a way to refactor code to so that there is no duplication. But its not possible to add a common helper lua file based on the swupdate design. We can only add one lua file per handler. So everything needs to be in that file. I discussed this with swupdate author and his reply was this is not supported.

recipes-extended/images/pelux-image-update/arp/emmcsetup1.lua
TSAnsari: sure this will reduce accidental errors. Will fix that.

PR #261, opened 2019-02-28
pelux.conf: set DISTRO_VERSION to 4.0

sashko:

### DIFF
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -1,6 +1,6 @@
 #
 #   Copyright (C) 2017 Pelagicore AB
-#   Copyright (C) 2018 Luxoft Sweden AB
+#   Copyright (C) 2018-2019 Luxoft Sweden AB
 #   SPDX-License-Identifier: MIT
 #
 
@@ -9,7 +9,7 @@ require conf/distro/poky.conf
 MAINTAINER = "Gordan Markuš <gordan.markus@pelagicore.com>"
 DISTRO = "pelux"
 DISTRO_NAME = "PELUX"
-DISTRO_VERSION = "3.0"
+DISTRO_VERSION = "4.0"
 TARGET_VENDOR = "-pelux"
 
 # Add distro features


conf/distro/pelux.conf
```diff
@@ -9,7 +9,7 @@ require conf/distro/poky.conf
 MAINTAINER = "Gordan Markuš <gordan.markus@pelagicore.com>"
```
thenor: We should remove the maintainer here also.

conf/distro/pelux.conf
```diff
@@ -9,7 +9,7 @@ require conf/distro/poky.conf
...
 DISTRO = "pelux"
 DISTRO_NAME = "PELUX"
-DISTRO_VERSION = "3.0"
+DISTRO_VERSION = "4.0"
```
thenor: I feel it is too early to change this version number.

sashko: why?

conf/distro/pelux.conf
sashko: yes, will do in a separate commit

sashko: actually, I believe this is an obligatory field and yocto-layer-check will fail, but let me check first...

sashko: nah, it's cool; will remove later.

conf/distro/pelux.conf
thenor: We're not on version 4. It's one thing for SWF to point to "4.0-alpha" but as far as I know this doesn't mention anything about a preliminary/alpha release.

sashko: of course, we're on 4.0 dev.

fhajredini:
Regarding the recent PR to go on 4.0, i disagree as well. This is the first time that i hear about 4.0 dev and so on.. We can hold this PR open until we release 4.0

fhajredini:
Are we keeping this PR until we release 4.0 or shall we close this?

sashko:
I strongly belive we should clearly distinguish released versions from the development ones. Perhaps setting the version to `4.0` just yet isn't the best idea, but the must be something that indicates this is not `3.0` because it is in fact not.

PR #263, opened 2019-03-03
core-image-pelux: add git to dev images

sashko:

### DIFF
--- a/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune-dev.bb
+++ b/layers/b2qt/recipes-core/images/core-image-pelux-qtauto-neptune-dev.bb
@@ -1,6 +1,6 @@
 #
 #   Copyright (C) 2017 Pelagicore AB
-#   Copyright (C) 2018 Luxoft Sweden AB
+#   Copyright (C) 2018-2019 Luxoft Sweden AB
 #   SPDX-License-Identifier: MIT
 #
 
@@ -12,6 +12,7 @@ require core-image-pelux-qtauto-neptune.bb
 IMAGE_FEATURES += "tools-debug tools-testapps ssh-server-openssh"
 IMAGE_INSTALL += "\
 	coreutils \
+	git \
 	openssh-sftp-server \
 	packagegroup-bistro-debug-utils \
 	vim \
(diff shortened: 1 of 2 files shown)


rpannek:
This question should have been answered in the commit message, because people in the future will have it too.

thenor:
The biggest reason for my skepticism is that we have been discussing that we want to make the images more lightweight and find what is included but not needed. If this is in fact needed for a specific use case, then that's one thing, but if this is just "would be nice to have" then I don't see any urgency and it can wait until a use case presents itself or at least until after the analysis of the image (an removal of unnecessary includes) is done.

rpannek:
Wait, so there was no discussion about it and it wasn't even attempted to be explained in the commit message? How in the world are people supposed to review something like that??

JEderonn:
As a developer, don't you develop on your own machine, using the SDK to cross-compile, and then deploy the artifacts to the target and test. (Or bitbake the complete image and deploy)

Has the team defined what a dev-image mean? I would assume that it has the same content as the non-dev image but with libraries with debug symbols, gdb, valgrind and ssh deployed to target. 

In the handbook it' stated: "Both PELUX images are available as -dev variants, which include some extra development and debugging tools. For some platforms, those images are also available as -update to generate update artifacts for the SOTA System."
Not so much, so perhaps there is a need to add more info about what a dev image is?

To summarize, I think there is a need to discuss this in the team before this pull request can be merged.

thenor:
> As a developer, don't you develop on your own machine, using the SDK to cross-compile, and then deploy the artifacts to the target and test. (Or bitbake the complete image and deploy)
> 
I personally don't see the need to develop on the Pelux image. I would develop on a separate machine and then recompile. I guess if you just want a specific package added on your Pelux image you could use scp.

> Has the team defined what a dev-image mean? I would assume that it has the same content as the non-dev image but with libraries with debug symbols, gdb, valgrind and ssh deployed to target.
> 
> In the handbook it' stated: "Both PELUX images are available as -dev variants, which include some extra development and debugging tools. For some platforms, those images are also available as -update to generate update artifacts for the SOTA System."
> Not so much, so perhaps there is a need to add more info about what a dev image is?
> 
> To summarize, I think there is a need to discuss this in the team before this pull request can be merged.
I agree that this most likely needs to be defined so that the entire team is aware of the purpose and use cases of the different images.

PR #265, opened 2019-03-04
avs-device-sdk: add recipe

sashko:
Added recipes for Alexa Voice Service (AVS) Software Development Kit.

Recipes were taken from https://github.com/alexa/avs-device-sdk and
adapted to PELUX.

### DIFF
--- a/recipes-alexa/avs-device-sdk/avs-device-sdk-1.9/0001-SpeechEncoder-Introduce-SpeechEncoder-with-OPUS-supp.patch
+++ b/recipes-alexa/avs-device-sdk/avs-device-sdk-1.9/0001-SpeechEncoder-Introduce-SpeechEncoder-with-OPUS-supp.patch
@@ -1,0 +1,843 @@
+From 9055b8dd917d27533d5bf6f1bcc7845720a44783 Mon Sep 17 00:00:00 2001
+From: Shotaro Uchida <shotaru@amazon.com>
+Date: Tue, 28 Aug 2018 13:51:42 -0700
+Subject: [PATCH 1/3] SpeechEncoder: Introduce SpeechEncoder with OPUS support
+
+---
+ CMakeLists.txt                                |   1 +
+ SpeechEncoder/CMakeLists.txt                  |   9 +
+ .../OpusEncoderContext/CMakeLists.txt         |   6 +
+ .../SpeechEncoder/OpusEncoderContext.h        |  61 +++++
+ .../OpusEncoderContext/src/CMakeLists.txt     |  30 +++
+ .../src/OpusEncoderContext.cpp                | 208 ++++++++++++++++++
+ .../include/SpeechEncoder/SpeechEncoder.h     |  88 ++++++++
+ SpeechEncoder/src/CMakeLists.txt              |  18 ++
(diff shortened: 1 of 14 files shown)


JEderonn:
I think we should use Alexa Auto SDK, not device SDK.

sashko:
> Alexa Auto SDK

Alexa Auto SDK is just a module for Alexa SDK.

JEderonn:
> Alexa Auto is just set of modules for Alexa SDK.
Ok, great, I didn't know, thought there were two variants of SDK.

martin-ejdestig:
Question: Is there some way we could use what is in the SDK directly without copying it? Or is that too hard/harder? Would be nice to not maintain copies, I am  guessing.

martin-ejdestig:
> Could you please rephrase your question? I am not sure I understand it correctly.

Use the recipes that are in the repository. You mentioned in #266 that the recipes only are for rocko. But I have no idea if you have tried to use them without copying since you just say "adapted to PELUX".

sashko:
> > Could you please rephrase your question? I am not sure I understand it correctly.
> 
> Use the recipes that are in the repository directly. You mentioned in #266 that the recipes only are for rocko. But I have no idea if you have tried to use them without copying since you just say "adapted to PELUX" without explaining why and what adaptations were necessary.

I cannot use the layer since Yocto since `sumo` requires `LAYERSERIES_COMPAT` to be set, and Amazon doesn't accept any PRs.

Also, it is not updated for a quite while, so we may end up bumping package version ourselves in the future. We need one recipe from the entire layer anyway.

erikboto:
Shouldn't recipes like this live in meta-bistro? According to the old thoughts, and the README of meta-pelux, this is the main idea behind meta-pelux and meta-bistro:

`The relationship with meta-bistro, which is also developed at Pelagicore, is that meta-bistro contains general software recipes and configurations, aimed at the IVI domain, but no distribution and no images, whereas meta-pelux builds on top of meta-bistro by adding a distro and recipes to build complete system images.`

sashko:
> Shouldn't recipes like this live in meta-bistro? According to the old thoughts, and the README of meta-pelux, this is the main idea behind meta-pelux and meta-bistro:
> 
> `The relationship with meta-bistro, which is also developed at Pelagicore, is that meta-bistro contains general software recipes and configurations, aimed at the IVI domain, but no distribution and no images, whereas meta-pelux builds on top of meta-bistro by adding a distro and recipes to build complete system images.`

Yeah, I've thought about the same. Just want it to go through the code review and then move them to `meta-bistro`.

erikboto:
> > Shouldn't recipes like this live in meta-bistro? According to the old thoughts, and the README of meta-pelux, this is the main idea behind meta-pelux and meta-bistro:
> > `The relationship with meta-bistro, which is also developed at Pelagicore, is that meta-bistro contains general software recipes and configurations, aimed at the IVI domain, but no distribution and no images, whereas meta-pelux builds on top of meta-bistro by adding a distro and recipes to build complete system images.`
> 
> Yeah, I've thought about the same. Just want it to go through the code review and then move them to `meta-bistro`.

Nice! I just had a quick look and it seems a few more recipes that should be in meta-bistro has sneaked in to meta-pelux, like mopidy and python-tornado.

sashko:
This will go to `meta-bistro` instead.

PR #266, opened 2019-03-04
aac-sdk: add recipes

sashko:
Added recipes for modules that implement Alexa Auto SDK interfaces.

### DIFF
--- a/recipes-alexa/aac-sdk/aac-sdk_1.4.0.bb
+++ b/recipes-alexa/aac-sdk/aac-sdk_1.4.0.bb
@@ -1,0 +1,19 @@
+SUMMARY = "Alexa Auto SDK"
+HOMEPAGE = "http://github.com/alexa/aac-sdk"
+SECTION = "libs"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+SRC_URI = "https://github.com/alexa/aac-sdk/archive/v${PV}.tar.gz"
+SRC_URI[md5sum] = "104fb98572def291f60dc95a5fb1eb77"
+SRC_URI[sha256sum] = "6dc631d194527a368ebd9916c7f2958f5d4d707d3d8ea0396bbd157f3f18d084"
+
+DEPENDS = "avs-device-sdk gtest"
+RDEPENDS_${PN} = "avs-device-sdk"
+
+inherit cmake pkgconfig
+
+S = "${WORKDIR}/aac-sdk-${PV}"
+
+FILES_SOLIBSDEV = ""
+SOLIBS = ".so"


martin-ejdestig:
This approach will clone the SDK repo multiple times?

sashko:
> This approach will clone the SDK repo multiple times?

No, just once as tarball will already exist in downloads. It'll unpack it separately to build each package. I don't know how to optimize it further without giving up modularity and simplicity of the recipes.

martin-ejdestig:
Well, main thing I was after is that it will be separate copies of the same source tree.

And also, will the same source be built multiple times? You point out subdirs but are there shared components?

recipes-alexa/aac-modules/aac-module-core_1.4.0.bb
```diff
@@ -0,0 +1,17 @@
...
+
+SUMMARY = "Alexa Automotive Core - Core Module"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://../../LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+DEPENDS = "avs-device-sdk sqlite3"
+
+S = "${WORKDIR}/aac-sdk-${PV}/modules/core/"
+
+inherit pkgconfig
```
DunderRoffe: Remove this inherit. It is already done in the include file

DunderRoffe:
Personally I think it is cleaner to do one recipe that provides all the packages. The main reason I think so is that you would only need one .bbappend file when overriding the SRURI instead of 5

recipes-alexa/aac-modules/aac-module-alexa_1.4.0.bb
```diff
@@ -0,0 +1,15 @@
+require aac-module.inc
+
+SUMMARY = "Alexa Automotive Core - Alexa Module"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://../../LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
```
DunderRoffe: Move LICENSE and LIC_FILES_CHKSUM to the include file

martin-ejdestig:
How are these libraries supposed to be used? There are no pkg-config files. Are they just modules to the avs-device-sdk added in #265 (that has a pkg-config file)? If they are to be used directly we need to patch in pkg-config files... ?

martin-ejdestig:
> Personally I think it is cleaner to do one recipe that provides all the packages. The main reason I think so is that you would only need one .bbappend file when overriding the SRURI instead of 5

It also looks like you get -dev and dpkg packages for each entry in PACKAGES . See:

https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#splitting-an-application-into-multiple-packages
https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#usingpoky-extend-customimage-customtasks

martin-ejdestig:
@sashko Hum, not obvious to me why you changed approach to a monolithic one. Too hard/much work with your first approach? No comment so I can only guess...

Also, have you seen: https://github.com/alexa/aac-sdk/tree/1.4/builder/meta-aac ? Do not know if it can be used but you have not mentioned it in comment or commit message so not sure if you know about it.

sashko:
> @sashko Hum, not obvious to me why you changed approach to a monolithic one. Too hard/much work with your first approach? No comment so I can only guess...

I thought it would be easier as otherwise, we would need to eiather split `-dev` packages as well, either ship all the headers even if we install one module only.

> Also, have you seen: https://github.com/alexa/aac-sdk/tree/1.4/builder/meta-aac ? Do not know if it can be used but you have not mentioned it in comment or commit message so not sure if you know about it.
> 
> Is it possible to specify a layer in a subdir of another repo? Means we would not have to maintain our own recipes, which would be nice, I guess. Don't know.

Yes, they've made a layer for `rocko`, which is not supported anymore. `Rocko` had older versions of some dependencies, so they've had to supply recipes for them as well. That's why they've had the whole layer.

Moreover, they do not have recipes for building the `Alexa Auto SDK`, but only `Alexa SDK` and dependancies.

martin-ejdestig:
> Moreover, they do not have recipes for building the `Alexa Auto SDK`, but only `Alexa SDK` and dependancies.

The link I posted is in the auto sdk... ?

sashko:
> > Moreover, they do not have recipes for building the `Alexa Auto SDK`, but only `Alexa SDK` and dependancies.
> 
> The link I posted is in the auto sdk... ?

Yes.

So, are you happy now or not?

martin-ejdestig:
> > > Moreover, they do not have recipes for building the `Alexa Auto SDK`, but only `Alexa SDK` and dependancies.
> > 
> > 
> > The link I posted is in the auto sdk... ?
> 
> Yes.
> 
> So, are you happy now or not?

If the answer is "Yes" then "they do not have recipes for building the `Alexa Auto SDK`" does not make any sense... or?

martin-ejdestig:
> So, are you happy now or not?

But yes, I am "happy" (removed the request for changes). I am guessing the recipes are too old here as well. But please be more forthcoming with details in you commit messages and comments in the future. (At least if you want me to review your prs.)

sashko:
> > > > Moreover, they do not have recipes for building the `Alexa Auto SDK`, but only `Alexa SDK` and dependancies.
> > > 
> > > 
> > > The link I posted is in the auto sdk... ?
> > 
> > 
> > Yes.
> 
> If the answer is "Yes" then the answer "they do not have recipes for building the `Alexa Auto SDK`" does not make any sense... or? (That is, it looks like there are recipes for the auto SDK as well.)

Yes, it does: they've provided a layer that builds `Alexa SDK`, which is a dependency of `Alexa Auto SDK`.

It's confusing, I understand. Just accept that we won't be easily able to use what Amazon provides us with "out of the box".

sashko:
This will go to `meta-bistro` instead.

PR #268, opened 2019-03-08
local.conf: set WIC_CREATE_EXTRA_ARGS to --no-fstab-update

TSAnsari:
A wks file used to create dual boot partitions was used to
automatically create an fstab file with mount entries for
the partitions. But since an fstab file already existed with
those entries, so it appended the new entries, resulting in
duplicate entries. Setting --no-fstab-update will ignore
generating the fstab file automatically.


martin-ejdestig:
Just a note. You can refer to the commit where this problem was introduced by adding its hash to the commit message. Something like:

"Manually created /etc/fstab was added in ae64f26bdf2488ff93b763118c5212598b956363 ."

Whole hash is not necessarily needed.

I also think that it would be better if we did not have a manually written /etc/fstab and just relied on the tool for generating it. But do not know how to fix the issues you had in #253 without digging in to it *much* more. :(

PR #269, opened 2019-03-13
core-image-pelux.bbclass: add alexa sdks

TSAnsari:
- avs-device-sdk
- aac-sdk
- portaudio-v19

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -26,6 +26,11 @@ IMAGE_INSTALL_append = "\
     node-state-manager \
 "
 
+# Alexa SDK
+IMAGE_INSTALL_append = "\
+    avs-device-sdk     \
+"
+
 # CAN utilities
 IMAGE_INSTALL_append = "\
     libsocketcan \


classes/core-image-pelux.bbclass
```diff
@@ -24,6 +24,9 @@ IMAGE_INSTALL_append = "\
     dlt-daemon         \
     dlt-daemon-systemd \
     node-state-manager \
+    portaudio-v19      \
```
sashko: portaudio is only needed for a test app, right?

also, why is this in GENIVI section?

martin-ejdestig: Yes, I am 99.99% sure. :) 

@TSAnsari Will there be a recipe for the test application? Is it not enough to add this dependency there? If there will not be a recipe, there must be some instructions for how to build the test app? If so, I'd say just add something like "bitbake portaudio-v19 and install the rpm:s" there. But... if there will be a recipe I would assume adding portaudio-v19 as a dependency in it would make this redundant.

martin-ejdestig: Hum, does the same go for the other dependencies as well? Should work by just having them as dependencies in relevant recipes? Does that make sense?

TSAnsari: > portaudio is only needed for a test app, right?
> 
> also, why is this in GENIVI section?

yes its only used by the sample application. Sure I will put it outside the GENIVI section.

TSAnsari: > portaudio is only needed for a test app, right?
> 
> also, why is this in GENIVI section?

So my thought was like this,
Since Sample application is part of the avs-device-sdk and is built along with it, we should install it along with avs-device-sdk. This is a temporary thing and should be removed once we develop our own alexa application/service. Then there will be separate recipe for our own application and depending on if we use portaudio or not, we can keep or remove it.

TSAnsari: > Yes, I am 99.99% sure. :)
> 
> @TSAnsari Will there be a recipe for the test application? Is it not enough to add this dependency there? If there will not be a recipe, there must be some instructions for how to build the test app? If so, I'd say just add something like "bitbake portaudio-v19 and install the rpm:s" there. But... if there will be a recipe I would assume adding portaudio-v19 as a dependency in it would make this redundant.


So my thought was like this,
Since Sample application is part of the avs-device-sdk and is built along with it, we should install it along with avs-device-sdk. This is a temporary thing and should be removed once we develop our own alexa application/service. Then there will be separate recipe for our own application and depending on if we use portaudio or not, we can keep or remove it.

sashko: I think we should package sampleapp separately then

TSAnsari: > I think we should package sampleapp separately then

Sample application is part of the sdk and is build along with the sdk, so that is the reason to add it here. We can add our own alexa service in separate recipe when we create one.

sashko: nothing stops you from packaging it separately

TSAnsari: Yes, but it feels unnecessary to create a separate recipe for sample app only to throw it away later. I remember we discussed this during standup and you also agreed on this. May be we can discuss this again tomorrow about what would be suitable approach.

Also this commit is not related to sample app. This commit only adds alexa sdks and portaudio to the image.

May be I should not add portaudio in this commit and use a different commit(along with commit for Sample app)

sashko: > ...it feels unnecessary to create a separate recipe ...
a package, not a recipe.

sashko: > May be I should not add portaudio
Yes, you shouldn't. Only sample app should depend on it.

TSAnsari: > > ...it feels unnecessary to create a separate recipe ...
> 
> a package, not a recipe.

Here is the PR that adds sample app
https://github.com/Pelagicore/meta-bistro/pull/219
Have a look and comment.

For this commit I will remove portaudio

TSAnsari: > > May be I should not add portaudio
> 
> No, you shouldn't. Only sample app should depend on it.

portaudio removed

thenor:
Disregard the Taha saying this doesn't build. The build is successful https://pelux.io/jenkins/job/meta-pelux/view/change-requests/job/PR-269/

thenor:
Typo in the commit message (two 'a' in packages).

sashko:
Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?

TSAnsari:
> Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?

We talked about removing portaudio in this commit. Will add it in a separate commit. So that was removed. Also removed aac-sdk as it is not required for current Alexa sample app.

sashko:
> > Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?
> 
> We talked about removing portaudio in this commit.

I do *not* recall such a conversation.

classes/core-image-pelux.bbclass
TSAnsari: > > > May be I should not add portaudio
> > 
> > 
> > No, you shouldn't. Only sample app should depend on it.
> 
> portaudio removed

Here

TSAnsari:
> > > Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?
> > 
> > 
> > We talked about removing portaudio in this commit.
> 
> I do _not_ recall such a conversation.



TSAnsari            May be I should not add portaudio

sashko        No, you shouldn't. Only sample app should depend on it.

TSAnsari    portaudio removed

sashko:
> > > > Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?
> > > 
> > > 
> > > We talked about removing portaudio in this commit.
> > 
> > 
> > I do _not_ recall such a conversation.
> 
> TSAnsari: May be I should not add portaudio
> 
> sashko: No, you shouldn't. Only sample app should depend on it.
> 
> TSAnsari: portaudio removed

you've also removed `aac-sdk` and the entire commit message.

TSAnsari:
> > > > > Hold on, @TSAnsari! What the heck? You've forced-pushed changes after complitely different change was reviewed?
> > > > 
> > > > 
> > > > We talked about removing portaudio in this commit.
> > > 
> > > 
> > > I do _not_ recall such a conversation.
> > 
> > 
> > TSAnsari: May be I should not add portaudio
> > sashko: No, you shouldn't. Only sample app should depend on it.
> > TSAnsari: portaudio removed
> 
> you've also removed `aac-sdk` and the entire commit message.

Yes the commit message had info about portaudio and aac-sdk. So it must be changed.

PR #274, opened 2019-03-18
Cherry-pick changes for Jenkinsfile to sumo

thenor:


sashko:
@thenor, should we do the same for `meta-bistro` and `meta-arp`? We don't really need it atm, but perhaps it would make sense not to leave them unaligned?

thenor:
> @thenor, should we do the same for `meta-bistro` and `meta-arp`? We don't really need it atm, but perhaps it would make sense not to leave them unaligned?

Yes. This should be done in all repos since the current setup in Jenkins only supports this way of writing the Jenkinsfile. I can cherry-pick, if you are doing other stuff.

PR #275, opened 2019-03-19
neptune3: Force to run as single process for rpi

sashko:
With neptune3, multiple applications in neptune ui
runs as separate processes. On rpi this causes some
performance issues since it doesn't have a very good
GPU. This causes a very low frame rate as well as
issues with correct rendering of the applications.
A workaround is added which runs neptune3 ui as a
single process. This fixes the performance and
rendering issues on rpi. This is a temporary fix,
until a solution is provided in neptune for this
issue.

(cherry picked from commit 1f81027db7c94c13b0cc1aeb03c86e2f4bc4db31)

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend
+++ b/layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend
@@ -11,6 +11,7 @@ RDEPENDS_${PN}_append = "\
       "
 
 HAS_CONTAINMENT = "${@bb.utils.contains('DISTRO_FEATURES', 'process-containment', '-c /opt/am/sc-config.yaml', '', d)}"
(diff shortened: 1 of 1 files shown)


sashko:
> > With neptune3, multiple applications in neptune ui
> > runs as separate processes. On rpi this causes some
> > performance issues since it doesn't have a very good
> > GPU. This causes a very low frame rate as well as
> > issues with correct rendering of the applications.
> > A workaround is added which runs neptune3 ui as a
> > single process. This fixes the performance and
> > rendering issues on rpi. This is a temporary fix,
> > until a solution is provided in neptune for this
> > issue.
> 
> It does not solve performance issue, it has not been confirmed. Do not mention that.
> Running in single-process is a workaround for a problem in the graphics driver on RPi. The symptom is that the widgets are not rendered properly.
> 
> Please update the commit messsage.

Alright, but this is just a cherry-pick.

sashko:
@JEderonn @TSAnsari ping.

PR #276, opened 2019-03-25
core-image-pelux.bbclass: add portaudio

TSAnsari:
portaudio and snowboy are required for Alexa sample
application to run. portaudio is used for microphone
input and snowboy implements wake word detector for
Alexa sample application.

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -28,6 +28,7 @@ IMAGE_INSTALL_append = "\
 
 # Alexa SDK
 IMAGE_INSTALL_append = "\
+    portaudio-v19      \
     avs-device-sdk     \
 "
 


classes/core-image-pelux.bbclass
```diff
@@ -28,6 +28,7 @@ IMAGE_INSTALL_append = "\
 
 # Alexa SDK
 IMAGE_INSTALL_append = "\
+    portaudio-v19      \
```
sashko: am I missing something? why it's not simply a dependency of the sdk?

TSAnsari: Since the portaudio has shared library. It has to be added to the image so that they are shipped with the image.

sashko: why?

TSAnsari: Because It is being used by the sample application.

TSAnsari: I am not sure but probably it will not be needed to explicitly add to the image if portaudio is set as an rdepends instead of depends. May be then it will implicitly add it to the image. I will try that and if it works then its not needed here. Will come back.

TSAnsari: I have tested now. So adding portaudio as rdepends will implicitly add it to the final image. Closing this PR.

TSAnsari:
Will add portaudio as an RDEPENDS which implicitly adds it to the final image. Closing this PR.

PR #281, opened 2019-03-29
swupdate: remove non-essential swupdate services

TSAnsari:
- swupdate-progress.service
- swupdate-usb@.service


sashko:
would have been nice to have explanations what they were about and why we didn't need them

martin-ejdestig:
> would have been nice to have explanations what they were about and why we didn't need them

Yes, always try to give an answer to "why?".

PR #284, opened 2019-04-01
linux-raspberrypi: enable cgroup memory

sashko:
Qt Application Manager parses /sys/fs/cgroup/memory/memory.stat to
obtain system memory usage information, which is later reported by
the diagnostics app of Neptune 3 UI.

Enable cgroup memory by adding "cgroup_enable=memory" to the command
line, so the app functions on Raspberry Pi.

Relevant information/discussion on using cgroup memory on RPi 3:
https://www.raspberrypi.org/forums/viewtopic.php?t=203128


JEderonn:
Approved

fhajredini:
Sign-off email is wrong

sashko:
> Sign-off email is wrong

Fixed it.

PR #286, opened 2019-04-05
systemd: remove network configuration

sashko:
PELUX uses connman as a network manager. Having this configuration in
place was resulting in system using two network management services at
the same time and obtaining 2 IP addresses from DHCP.


fhajredini:
Approved!

fhajredini:
Would be nice to test this before merging tho

sashko:
> Would be nice to test this before merging tho

This was of course tested. By both me and Ekaterina.

sashko:
> Approved!

Forgot how to use github? :rofl:

PR #287, opened 2019-04-05
core-image-pelux: remove obsolete cmake sdk fix

sashko:
The issue appears to be fixed in the mainstream.


sashko:
This code is a few years old and was added without any explanations.

Google search has shown that cmake config for Yocto SDK had been broken for a while, but it's not clear to me which commit has fixed it, perhaps https://patchwork.openembedded.org/patch/152556/.

sashko:
Perhaps @GordanM or @jonte remember what it was about.

PR #289, opened 2019-04-07
Add Arabic font

sashko:


Gagi2k:
Looks good to me, although i would prefer this to go into meta-boot2qt directly

sashko:
> Looks good to me, although i would prefer this to go into meta-boot2qt directly

I agree - it has to be upstreamed.

sashko:
https://codereview.qt-project.org/#/c/258376/

PR #291, opened 2019-04-10
neptune.service: Workaround for webengine

fhajredini:
Webapps such as youtube are failing on both NUC and RPI.
This environment line will be a workaround fix to both.

### DIFF
--- a/layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune.service
+++ b/layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune.service
@@ -12,6 +12,7 @@ Environment=LC_ALL=en_US
 Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/root/dbus/user_bus_socket
 Environment=QT_IM_MODULE=qtvirtualkeyboard
 Environment=XDG_RUNTIME_DIR=/tmp
+Environment=QTWEBENGINE_CHROMIUM_FLAGS="--no-sandbox --single-process"
 
 [Install]
 WantedBy=multi-user.target


martin-ejdestig:
Please explain more in the commit message. (What is the error? Log excerpts? Why does running in single process help? Can it be reverted once something is fixed? Are there bugs you could refer to, etc. etc.)

layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune.service
```diff
@@ -12,6 +12,7 @@ Environment=LC_ALL=en_US
 Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/root/dbus/user_bus_socket
 Environment=QT_IM_MODULE=qtvirtualkeyboard
 Environment=XDG_RUNTIME_DIR=/tmp
+Environment=QTWEBENGINE_CHROMIUM_FLAGS="--no-sandbox --single-process"
```
TSAnsari: If --single-process is used for all targets then remove its inclusion for only rpi in 
https://github.com/Pelagicore/meta-pelux/blob/master/layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend

Remove 
FORCE_SINGLE_PROCESS = "${@bb.utils.contains('MACHINE', 'raspberrypi3', ' --force-single-process', '', d)}" line and fix the recipe accordingly.
unless the --single-process for QTWEBENGINE_CHROMIUM_FLAGS works differently from the --single-process commandline argument for neptune-ui

sashko:
`neptune.service: run qtwebengine in single process mode` or whatever

layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune.service
fhajredini: From what i have tested.. the `QTWEBENGINE_CHROMIUM_FLAGS` environment variable, and the actual neptune service arguments are different from each other. Therefore the `--force-single-process` and `QTWEBENGINE_CHROMIUM_FLAGS="--single-process"` are different.

martin-ejdestig:
> There has been no errors specified besides the application exiting itself.

There is no error output or you have not looked?

> Once the rootcause is fixed this can be reverted.

Is it tracked somewhere? If yes, where? If not, should it not be?

layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune.service
JEderonn: Can confirm what @fhajredini states.

martin-ejdestig: This (that they are different and that we do X for rpi alredy) is a perfect example of something worth mentioning in the commit message IMHO. Would:

1.  Make it easier for reviewers. Tariq would not have had to ask about it. Others who did not know about it are informed.

2. Make it more clear what is going on when looking at this in the future. People will not have to relearn the hard way what you already learned. Will even future Fisnik remember the details about this a year from now?

Please update the commit message and try to think more about putting useful information like this in future commit messages.

(Would also be nice to have more details as to why they are different. Might require looking at e.g. the Neptune code though... do not have to go there... I guess...)

/Commi message police signing off. :)

DunderRoffe:
I agree with @martin-ejdestig. This workaround have been added to demos before as a quick fix, but if this is going to be added to mainline PELUX this should be traced so the root cause can be known and fixed.

martin-ejdestig:
neptune: run qtwebengine in single process mode

QtWebEngine crashes trying load some websites, e.g. youtube.com.

This is an issue of the QtWebEngine. Running it outside the sandbox
and in the single process mode is a temporary workaround.

fhajredini:
Created a bug and added it on the commit message.

martin-ejdestig:
Still no mention of what was discussed in Tariq's issue in commit message.

martin-ejdestig:
I would change:

```
The QTWEBENGINE_CHROMIUM_FLAGS environment variable, and the actual
neptune.service arguments are different from each other. Therefore the
--force-single-process (used as an argument when starting neptune) and
QTWEBENGINE_CHROMIUM_FLAGS="--single-process" are different. The fix is
not required on the neptune itself, its for the QTWEBENGINE.
```

to something like:

```
--single-process is already passed to the Neptune executable itself for raspberrypi.
See FORCE_SINGLE_PROCESS in layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend .
The command line flag is not enough to fix e.g. Youtube crashing though.
```

Main thing I am after is that you refer more clearly to the already existing rpi workaround.

(Is the rpi specific workaround still needed with this... ? *sigh* Anyway, just make the way you refer to the other fix more clear.)

PR #293, opened 2019-04-12
qtivi-mopidy-plugin: add recipe

sashko:
This recipe builds and installs Mopidy media backend plugin for Qt IVI.


recipes-multimedia/qtivi-mopidy-plugin/qtivi-mopidy-plugin_git.bb
```diff
@@ -0,0 +1,22 @@
+SUMMARY = "Qt IVI Mopidy media playback plugin"
+HOMEPAGE = "https://github.com/Pelagicore/qtivi-qtivi-plugin"
+SECTION = "libs"
```
JEderonn: Consider change section to "libs/multimedia"

PR #295, opened 2019-04-16
grub-efi: fix file permissions

sashko:
All service files should have 644 permissions.

Permissions of grubenv on Ubuntu is also 644, so do the same on PELUX.

### DIFF
--- a/meta-intel-extras/recipes-support/grub/grub-efi_%.bbappend
+++ b/meta-intel-extras/recipes-support/grub/grub-efi_%.bbappend
@@ -13,9 +13,9 @@ do_install_append() {
     install -d ${D}${datadir}
     install -d ${D}${systemd_system_unitdir}
 
-    install -m 0755 ${WORKDIR}/grubenv ${D}${datadir}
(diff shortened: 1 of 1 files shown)


meta-intel-extras/recipes-support/grub/grub-efi_%.bbappend
```diff
@@ -13,9 +13,9 @@ do_install_append() {
     install -d ${D}${datadir}
     install -d ${D}${systemd_system_unitdir}
 
-    install -m 0755 ${WORKDIR}/grubenv ${D}${datadir}
+    install -m 0644 ${WORKDIR}/grubenv ${D}${datadir}
     install -m 0755 ${WORKDIR}/grubenv-copy.sh ${D}${bindir}
```
TSAnsari: grubenv-copy.sh also needs the same permissions

sashko: It is a executable, so no, it has to have x permission set.

PR #298, opened 2019-04-25
Jenkinsfile: script for previous job interuption

pookimon:
Script will abort previous running build if new build is in progress. It will be helpful especially for Pull Requests builds


fhajredini:
Include a few more details on the commit message

pookimon:
> Better explanation please: why is this needed.

Script will abort previous running build if new build is in progress.
It will be helpful especially for Pull Requests builds to not waste resources

Could you please suggest what else can be added.

PR #301, opened 2019-04-26
Jenkinsfile: script for build abortion moved on top of node assignment

pookimon:
Script should be moved on top of node assignment as previous build will not be interrupted when all nodes are busy
Signed-off-by: Dmytro Iurchuk <diurchuk@luxoft.com>


fhajredini:
Improve the commit message by adding the comment that you added here: 
"Script should be moved on top of node assignment as previous build will not be interrupted when all nodes are busy"

PR #302, opened 2019-05-02
Use PARTUUID to boot intel and ARP targets

TSAnsari:
NUC and ARP were assumed to boot from /dev/sda.
However, the boot device could be other than the
usb. In that case hardcoded device /dev/sda would
fail to boot. The solution is to use manually
generated PARTUUIDs hardcoded in wks file instead
of hardcoded device names, which could change. The
PARTUUIDs will remain the same even if the boot
device is changed. This also affected the swupdate
which also assumed the device to be /dev/sda.
Software update is modifed to use PARTLABEL to
detect the boot device. Since PARTUUID and
PARTLABEL have one to one mapping, PARTUUID
can be used in software update scripts. However,
for better readability reasons PARTLABEL are used.

### DIFF
--- a/recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup2.lua
+++ b/recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup2.lua
@@ -5,11 +5,6 @@ function os.capture(cmd)
 	return s
 end
 
-function file_exists(name)
-	local f=io.open(name,"r")
-	if f~=nil then io.close(f) return true else return false end
-end
-
 function cmdexec(cmd)
 	local ret, s, status = os.execute(cmd)
 	if (status ~= 0) then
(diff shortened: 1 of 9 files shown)


TSAnsari:
> Is it really possible to use PARTUUID? Have you tried to use the image on another hardware setup, using another USB stick? I'm not an expert but isn't the PARTUUID unique for the device and the partitioning that device has?

The PARTUUID is not dynamically generated here. It is generated once manually and hardcoded in the wks file. In a way it works as a fixed partition label rather than a UUID.

JEderonn:
> > Is it really possible to use PARTUUID? Have you tried to use the image on another hardware setup, using another USB stick? I'm not an expert but isn't the PARTUUID unique for the device and the partitioning that device has?
> 
> The PARTUUID is not dynamically generated here. It is generated once manually and hardcoded in the wks file. In a way it works as a fixed partition label rather than a UUID.

Ok, thanks for clarifying. Perhaps it's worth to mention it somewhere.

martin-ejdestig:
(I have not done a thorough review, but thought I'd mention this here lest I forget.)

It is not clear to me why you need to use partition UUID:s in `fstab` and labels for software update. You mention that this is the case in the commit message but not why. (It does not work? It is not supported? Or... ?)

TSAnsari:
> (I have not done a thorough review, but thought I'd mention this here lest I forget.)
> 
> It is not clear to me why you need to use partition UUID:s in `fstab` and labels for software update. You mention that this is the case in the commit message but not why. (It does not work? It is not supported? Or... ?)

I could have used PARTUUID in swupdate scripts and it would have worked the same way as label. The only reason I used label was that it is more readable in the script. uuid and label have one to one mapping in the wks file.

martin-ejdestig:
> I used label was that it is more readable in the script.

Check. Just mention that so future @TSAnsari and me also can find out by checking commit message. :)

TSAnsari:
> > > Is it really possible to use PARTUUID? Have you tried to use the image on another hardware setup, using another USB stick? I'm not an expert but isn't the PARTUUID unique for the device and the partitioning that device has?
> > 
> > 
> > The PARTUUID is not dynamically generated here. It is generated once manually and hardcoded in the wks file. In a way it works as a fixed partition label rather than a UUID.
> 
> Ok, thanks for clarifying. Perhaps it's worth to mention it somewhere.

Will update the commit msg

TSAnsari:
> > I used label was that it is more readable in the script.
> 
> Check. Just mention that so future @TSAnsari and me also can find out by checking commit message. :)

Will update the commit msg

TSAnsari:
Updated commit msg

TSAnsari:
@martin-ejdestig and @JEderonn If there are no more comments then I would like to merge this change. Have already updated the commit message according to previous comments.

recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup2.lua
```diff
@@ -19,109 +14,113 @@ function cmdexec(cmd)
...
+	local ret
+	local updatedev
+
+	ret, updatedev = get_update_device_from_boot_device_label("platform1")
+	if (ret == false) then
+		return ret, updatedev .. ", Cannot find update device"
+	end
+
+	local mountpoint = "/tmp/mounted" .. updatedev
+	local mountdev = "/dev/" .. updatedev
```
martin-ejdestig: Perhaps explain that /dev/updatedev is needed due to it not being possible to use labels or UUID:s in sw-description?

martin-ejdestig:
Too bad that exact same files have to be duplicated. Found this diff:

```
--- recipes-extended/images/pelux-image-update/arp/emmcsetup2.lua	2019-03-19 14:08:55.971691230 +0100
+++ recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup2.lua	2019-03-19 14:08:55.971691230 +0100
@@ -78,7 +78,7 @@ function preinst()
 	f:close()
 
 	if (ret == nil) then
-		return ret, out
+		return false, out
 	end
 
 	out = os.capture("/usr/sbin/sfdisk --force " .. eMMC .. " < /tmp/partitions")
@@ -87,7 +87,7 @@ function preinst()
 
 	ret, out = cmdexec("/usr/sbin/partprobe " .. eMMC)
 	if (false == ret) then
-		return false, out
+		return ret, out
 	end
 
 	return true, out
```

TSAnsari:
> Too bad that exact same files have to be duplicated. Found this diff:
> 
> ```
> --- recipes-extended/images/pelux-image-update/arp/emmcsetup2.lua	2019-03-19 14:08:55.971691230 +0100
> +++ recipes-extended/images/pelux-image-update/intel-corei7-64/emmcsetup2.lua	2019-03-19 14:08:55.971691230 +0100
> @@ -78,7 +78,7 @@ function preinst()
>  	f:close()
>  
>  	if (ret == nil) then
> -		return ret, out
> +		return false, out
>  	end
>  
>  	out = os.capture("/usr/sbin/sfdisk --force " .. eMMC .. " < /tmp/partitions")
> @@ -87,7 +87,7 @@ function preinst()
>  
>  	ret, out = cmdexec("/usr/sbin/partprobe " .. eMMC)
>  	if (false == ret) then
> -		return false, out
> +		return ret, out
>  	end
>  
>  	return true, out
> ```

unfortunately there is no way to avoid this duplication. I asked the author of swupdate regarding this and the answer was that it is not possible.

martin-ejdestig:
> unfortunately there is no way to avoid this duplication. I asked the author of swupdate regarding this and the answer was that it is not possible.

OK. But fine now. There used to be a difference but I diffed it wrongly now. :)

A bit off topic but perhaps it would be possible to have symlinks to the scripts? (Not asking for this to be done now, just something we could do to avoid copying perhaps.)

PR #303, opened 2019-05-07
softwarecontainer: run neptune Map app in container

TSAnsari:
This runs neptune Map application in softwarecontainer.
It requires networkgateway to be configured in order to
access api.mapbox.com API which provides map information
to render. Filegateway is also configured since the
iptables used by networkgateway require to lock
/run/xtables.lock when running commands. /etc/ssl is
also mounted through filegateway. This is needed for SSL
certificates needed to access https://api.mapbox.com.

### DIFF
--- a/recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
+++ b/recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
@@ -36,6 +36,40 @@
           ]
         },
         {
+          "id": "file",
+          "config": [
+            {
+              "path-host": "/run",
+              "path-container": "/run",
+              "read-only": true
+            },
+            {
+              "path-host": "/etc/ssl",
+              "path-container": "/etc/ssl",
+              "read-only": true
+            }
+          ]
+        },
+        {
+          "id":"network",
+          "config": [
+            {
+              "direction": "OUTGOING",
+              "allow": [
+                { "host": "*", "protocols": ["udp", "tcp"], "ports": 53 },
+                { "host": "*", "protocols": "tcp", "ports": [80, 443] }
+              ]
+            },
+            {
+              "direction": "INCOMING",
+              "allow": [
+                { "host": "*", "protocols": ["udp", "tcp"], "ports": 53 },
+                { "host": "*", "protocols": "tcp", "ports": [80, 443] }
+              ]
+            }
+          ]
+        },
+        {
           "id": "devicenode",
           "config": [
             {
--- a/layers/b2qt/recipes-qt/automotive/qtapplicationmanager/sc-config.yaml
+++ b/layers/b2qt/recipes-qt/automotive/qtapplicationmanager/sc-config.yaml
@@ -6,6 +6,7 @@ formatType: am-configuration
 ---
 containers:
   selection:
+    - com.pelagicore.map: "softwarecontainer"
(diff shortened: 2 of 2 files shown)


recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
```diff
@@ -35,6 +35,38 @@
...
+          "config": [
+            {
+              "path-host": "/run",
+              "path-container": "/run",
+              "read-only": false
+            },
+            {
+              "path-host": "/etc/ssl",
+              "path-container": "/etc/ssl",
+              "read-only": false
```
sashko: are you sure ro wouldn't be enough?

recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
```diff
@@ -35,6 +35,38 @@
...
+            }
+          ]
+        },
+        {
+          "id":"network",
+          "config": [
+            {
+              "direction": "OUTGOING",
+              "allow": [
+                { "host": "*" }
```
sashko: wouldn't it be better to allow traffic to map servers only?

TSAnsari: I missed some change here. I will update this.

recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
TSAnsari: Probably. Will try read-only true

recipes-containers/softwarecontainer/files/io.qt.ApplicationManager.Application.json
TSAnsari: I tried using api.mapbox.com as hostname but that it is dropping some traffic resulting in failing to render map correctly. Probably map application is accessing something more than map server. So will keep the host as *

sashko: or use Wireshark to check what hosts it is accessing...

JEderonn: Have you checked https://docs.mapbox.com/help/troubleshooting/firewalls/ ? Add the list of domains and see if it works.

TSAnsari: Thanks @JEderonn I will try this

TSAnsari: Whitelisting these domains doesn't work. There is still a lot of traffic being dropped and hence no proper rendering of Map. 
I suggest that for the coming release this week, we can allow all traffic to get the maps working. It would have been a good example to selectively white list the hosts, but that can be investigated later.

JEderonn: Ok, I would then propose that you create a new story for it and refer to the issue # here for tracking.

TSAnsari: User story created 
https://luxproject.luxoft.com/jira/browse/OAI1209-585

PR #308, opened 2019-05-16
core-image-pelux: add connectivity-manager

martin-ejdestig:
We should probably not keep adding components unconditionally like this to core-image-pelux*.bbclass . Do it for connectivity-manager for now and then we need to come back and think about how we should handle "features" at some later point (soon (tm)).

Depends on https://github.com/Pelagicore/meta-bistro/pull/246 .

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -52,6 +52,11 @@ IMAGE_INSTALL_remove_qemux86-64 = "\
     swupdate \
 "
 
+# Connectivity Manager
+IMAGE_INSTALL_append = "\
+    connectivity-manager \
+"
+
 IMAGE_INSTALL_append_arp = "\
     arp-driver \
 "


classes/core-image-pelux.bbclass
```diff
@@ -52,6 +52,11 @@ IMAGE_INSTALL_remove_qemux86-64 = "\
     swupdate \
 "
 
+# Connectivity Manager
+IMAGE_INSTALL_append = "\
```
TSAnsari: Will it be possible to test connectivity-manager on qemu? If not then may be remove from qemu.

martin-ejdestig: To test WiFi itself one would have to set that up in QEMU. Do not know how to do that. But should still be possible to run it. And if we add e.g. ethernet that should be possible to test.

martin-ejdestig:
> commit message is not your personal thoughts journal

Not good to mention that it is something we should not really do? Do you want me to reword it (if so, please tell me what to write) or remove it completely so that this only says "add recipe"?

martin-ejdestig:
Changed commit message from:

```
 core-image-pelux: add connectivity-manager

We should probably not keep adding components unconditionally like this
to core-image-pelux*.bbclass . Do it for connectivity-manager for now and
then we need to come back and think about how we should handle "features"
at some later point (soon (tm)).
```

to

```
core-image-pelux: add connectivity-manager
    
Want it (https://github.com/Pelagicore/connectivity-manager) in today's
nightly and the next weekly so it can be tested in time for 4.0 release
sometime next week.
    
Not sure that adding individual components like this unconditionally to
core-image-pelux*.bbclass is something we should keep doing. Should
probably look into "distro features" (see e.g. the process-containment
feature) for the new components we are building and what we already have
added in core-image-pelux*.bbclass sometime soon.
```

martin-ejdestig:
@sashko Do you want me to remove mention of distro features altogether?

PR #310, opened 2019-05-17
core-image-pelux-qtauto: add qttools-tools

sashko:
qttools-tools contains qtdiag app which is used by the Neptune's
Diagnostic application.


JEderonn:
Consider add tools to the dev-image only, or explain why you think there is a need to have it installed on every QtAS-image?

Gagi2k:
I think he is right that it is needed, as we have a info screen in neptune3 which usese qtdiags output to show information about the system which helps us to identify problems, but instead of adding it to the image recipes, add it as a runtime dependency to neptune3 and upstream it.

PR #314, opened 2019-05-27
Media partition

fhajredini:


scripts/lib/wic/canned-wks/mkefidisk-multiboot.wks
```diff
@@ -10,4 +10,6 @@ part / --source rootfs --ondisk sda --fstype=ext4 --label platform --align 1024
 
 part swap --ondisk sda --size 44 --label swap1 --fstype=swap
 
+part /media --ondisk sda --size 787 --fstype=ext4 --label media --align 1024
```
TSAnsari: Why 787? why not 1 GB? If you are doing this due to overhead factor, then it will be good to explicitly use the overhead option so anyone looking at this file can understand the reasoning behind this particular size. I think it will be better to use 1 GB directly and set the factor to 1.

scripts/lib/wic/canned-wks/mkefidisk-multiboot.wks
```diff
@@ -4,9 +4,9 @@
 
 part /boot --source bootimg-efi --sourceparams="loader=grub-efi" --ondisk sda --label msdos --active --align 1024
 
-part / --source rootfs --ondisk sda --fstype=ext4 --label platform --align 1024
+part / --source rootfs --ondisk sda --fstype=ext4 --label platform --size 2756 --align 1024
```
TSAnsari: Again would be better to use the overhead factor explicitly here. So the one knows why the particular size is chosen. 
Better would be to set the --size to the final size and set overhead factor to 1

TSAnsari: Also you may get conflict when you try to merge this as this wks file has changed. So first rebase.

meta-intel-extras/recipes-core/base-files/base-files/intel-corei7-64/fstab
```diff
@@ -1,3 +1,4 @@
 /dev/root            /                    auto       defaults              1  1
 /dev/sda1	/boot	vfat	defaults	0	0
```
TSAnsari: The /dev/sda1 and /dev/sda4 are replaced by PARTUUID's so generate PARTUUID for sda5 also.

scripts/lib/wic/canned-wks/mkefidisk-multiboot.wks
TSAnsari: generate PARTUUID for this partition. Once you rebase you will see the changes and how to fix the PARTUUID for media partition.

fhajredini: When i set 787 i expected the multiplier by 1.3 (so that it reaches 1 GB) but that didnt happen for this partition. Currently the image size is 7.9 GB and if i set this partition  to 1 GB, then it will be ~8.2 GB. Since we changed the requirements to a 16GB USB, then i'll change this.

fhajredini:
The rebase has been done. The only thing left from @TSAnsari comments is the size of the media partition which i left it on 787 MB. The only reason i left it on 787, is to try and remain under 8GB. At this point, im not sure if that is a valid reason or not.

scripts/lib/wic/canned-wks/mkefidisk-multiboot.wks
fhajredini: The size factor has been mentioned on the commit messages, you think its still good to add it there? Moreover, explicitly set it to 1?

TSAnsari:
> The rebase has been done. The only thing left from @TSAnsari comments is the size of the media partition which i left it on 787 MB. The only reason i left it on 787, is to try and remain under 8GB. At this point, im not sure if that is a valid reason or not.

I my opinion its better to explicitly set the size without any assumption of overhead factor. This makes the intention clear. So either set the size to 787 and explicitly use overhead factor of 1.3 or set size to one GB and overhead factor to 1.

Perhaps others may have different opinion. @JEderonn @martin-ejdestig @sashko

PR #315, opened 2019-05-28
Jenkinsfile: Vagrant to docker migration

pookimon:
In matter of meta layers replaceLayer mechanism was changed. Now meta repository dir
will be mounted within container, and all other action will take place inside. Everything
else was inerhited from pelux-manifest

Signed-off-by: Dmytro Iurchuk <diurchuk@luxoft.com>

### DIFF
--- a/scripts/Jenkinsfile
+++ b/scripts/Jenkinsfile
@@ -23,10 +23,15 @@ void buildPelux(String variant, String imageName) {
         sh "rm -rf pelux-manifests/"
         sh "git clone https://github.com/Pelagicore/pelux-manifests.git -b ${branchName}"
         println("Using the specified branch for build: ${branchName}")
+        sh "cp -R ${peluxDir} pelux-manifests/${peluxDir}"
         dir('pelux-manifests') {
-            def code = load "ci-scripts/yocto.groovy"
+            def customImage = docker.image("pelux/pelux-yocto:yoctouser")
             String peluxPath = "${env.WORKSPACE}/${peluxDir}"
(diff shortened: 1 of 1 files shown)


sashko:
`Jenkins: ...`

scripts/Jenkinsfile
```diff
@@ -23,10 +23,15 @@ void buildPelux(String variant, String imageName) {
         sh "rm -rf pelux-manifests/"
         sh "git clone https://github.com/Pelagicore/pelux-manifests.git -b ${branchName}"
         println("Using the specified branch for build: ${branchName}")
+        sh "cp -R ${peluxDir} pelux-manifests/${peluxDir}"
```
martin-ejdestig: Hum, is this copy needed? Could you not bind mount the workspace dir instead of $WORKSPACE/pelux-manifests? If I remember correctly, copying is done in yocto.groovy as well.

pookimon: It was moved from yocto.groovy here, as all other operations will be performed inside container

martin-ejdestig: OK, I see now that you have changed it from a `cp` in `yocto.groovy` to a `mv` in `yocto2.groovy`. (I did not have time to look at https://github.com/Pelagicore/pelux-manifests/pull/371 yesterday before it was merged.)

martin-ejdestig: I guess we can tweak things like this later if we want to. Resolving.

pookimon: I apologize

PR #316, opened 2019-06-03
rpi partitioning: create dual and media partitions

TSAnsari:
- creates dual rootfs partitions of approx 1.5 GB each taking into
  account the overhead factor of 1.3. The dual partitioning is done
  to facilitate swupdate feature which use A/B partitioning scheme.

- creates media partition of size 1 GB.

### DIFF
--- a/meta-rpi-extras/classes/dualpart_image-rpi.bbclass
+++ b/meta-rpi-extras/classes/dualpart_image-rpi.bbclass
@@ -1,0 +1,125 @@
+#   Copyright (c) 2012-2019 meta-raspberrypi contributors
+#
+#   This file was copied from meta-raspberrypi page and only has minor changes.
+#   They include a new data partition and set static partition sizes.
+#
+#   The original file can be found here: https://github.com/agherzan/meta-raspberrypi/blob/master/classes/sdcard_image-rpi.bbclass
+#
+#   SPDX-License-Identifier: MIT
+
+inherit sdcard_image-rpi
+
+IMAGE_CMD_rpi-sdimg () {
+
+    # Align partitions
+    ROOTFS_SIZE=1490944
+    DATA_SIZE=1048576
+    BOOT_SPACE_ALIGNED=$(expr ${BOOT_SPACE} + ${IMAGE_ROOTFS_ALIGNMENT} - 1)
+    BOOT_SPACE_ALIGNED=$(expr ${BOOT_SPACE_ALIGNED} - ${BOOT_SPACE_ALIGNED} % ${IMAGE_ROOTFS_ALIGNMENT})
+    SDIMG_SIZE=$(expr ${IMAGE_ROOTFS_ALIGNMENT} + ${BOOT_SPACE_ALIGNED} + $ROOTFS_SIZE + $ROOTFS_SIZE + $DATA_SIZE)
+
+    echo "Creating filesystem with Boot partition ${BOOT_SPACE_ALIGNED} KiB and RootFS $ROOTFS_SIZE KiB"
+
+    # Check if we are building with device tree support
+    DTS="${KERNEL_DEVICETREE}"
+
+    # Initialize sdcard image file
+    dd if=/dev/zero of=${SDIMG} bs=1024 count=0 seek=${SDIMG_SIZE}
+
+    dd if=/dev/zero of=${IMGDEPLOYDIR}/media_partition.ext3 bs=1024 count=0 seek=$DATA_SIZE
+
+    # Create partition table
+    parted -s ${SDIMG} mklabel msdos
+    # Create boot partition and mark it as bootable
+    parted -s ${SDIMG} unit KiB mkpart primary fat32 ${IMAGE_ROOTFS_ALIGNMENT} $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT})
+    parted -s ${SDIMG} set 1 boot on
+    # Create rootfs partition to the end of disk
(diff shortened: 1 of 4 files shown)


TSAnsari:
The IMAGE_CMD_rpi-sdimg() function from sdcard_image-rpi.bbclass is overidden in dualpart_image-rpi.bbclass. Bulk of the code in this function is same as in the base class. Some code is added and updated to create dual rootfs partitions and media partition. Also the variable ROOTFS_SIZE is overridden with fixed rootfs size.

meta-rpi-extras/classes/dualpart_image-rpi.bbclass
```diff
@@ -0,0 +1,116 @@
+inherit sdcard_image-rpi
```
martin-ejdestig: A lot of copying here. I have not looked at details yet... if it is not possible to extend `meta-raspberrypi/classes/sdcard_image-rpi.bbclass` we must at least add a header with copyright and license information.

I assume it is hard/impossible to extend? If so, perhaps a blurb at the top about what changes have been made as well and where the original function comes from.

TSAnsari: yes its not possible to just extend by appending. The required changes are scattered in the function, so unfortunately have to override and copy everything + updates/additions. Will add comments in the code to explain this.
Yes, I was also thinking about the license thing.

meta-rpi-extras/classes/dualpart_image-rpi.bbclass
```diff
@@ -0,0 +1,116 @@
...
+    # Create boot partition and mark it as bootable
+    parted -s ${SDIMG} unit KiB mkpart primary fat32 ${IMAGE_ROOTFS_ALIGNMENT} $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT})
+    parted -s ${SDIMG} set 1 boot on
+    # Create rootfs partition to the end of disk
+    parted -s ${SDIMG} -- unit KiB mkpart primary ext2 $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT}) $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT} \+ $ROOTFS_SIZE)
+    parted -s ${SDIMG} -- unit KiB mkpart primary ext2 $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT} \+ $ROOTFS_SIZE) $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT} \+ $ROOTFS_SIZE \+ $ROOTFS_SIZE)
+    parted -s ${SDIMG} -- unit KiB mkpart primary ext3 $(expr ${BOOT_SPACE_ALIGNED} \+ ${IMAGE_ROOTFS_ALIGNMENT} \+ $ROOTFS_SIZE \+ $ROOTFS_SIZE) -1s
+    parted ${SDIMG} print
+
+    mkfs.ext3 -F ${IMGDEPLOYDIR}/media_partition.ext3 -b 1024
```
martin-ejdestig: Why set the block size for the filesystem to 1024 here? I can not see that meta-raspberrypi does that for the root fs. In fact... mkfs is not called at all for the rootfs? How does this work? `parted mkpart` should create the file system... ?

meta-rpi-extras/classes/dualpart_image-rpi.bbclass
```diff
@@ -0,0 +1,116 @@
...
+        xzcat ${SDIMG_ROOTFS} | dd of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024)
+        xzcat ${SDIMG_ROOTFS} | dd of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024 + $ROOTFS_SIZE \* 1024)
+        xzcat ${IMGDEPLOYDIR}/media_partition.ext3 | dd of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024 + $ROOTFS_SIZE \* 1024 + $ROOTFS_SIZE \* 1024)
+    else
+        dd if=${SDIMG_ROOTFS} of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024)
+        dd if=${SDIMG_ROOTFS} of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024 + $ROOTFS_SIZE \* 1024)
+        dd if=${IMGDEPLOYDIR}/media_partition.ext3 of=${SDIMG} conv=notrunc seek=1 bs=$(expr 1024 \* ${BOOT_SPACE_ALIGNED} + ${IMAGE_ROOTFS_ALIGNMENT} \* 1024 + $ROOTFS_SIZE \* 1024 + $ROOTFS_SIZE \* 1024)
+    fi
+
+    rm -rf ${IMGDEPLOYDIR}/media_partition.ext3
```
martin-ejdestig: Not that important... so feel free to ignore... but perhaps check if this file is sparse? If it is, no need to rm to save space... minimizes diff. :)

fhajredini:
The image has been moved from rpi-sdimg to wic with #334. The media partition was added there as well.

PR #322, opened 2019-06-13
Arp partitioning

fhajredini:
meta-arp-extras: new fstab for ARP 
conf/variant: include meta-arp-extras

### DIFF
--- a/conf/variant/arp/bblayers.conf.sample
+++ b/conf/variant/arp/bblayers.conf.sample
@@ -6,6 +6,7 @@ require conf/variant/common/bblayers.conf
 
 BBLAYERS += "\
     ${BSPDIR}/sources/meta-pelux/meta-intel-extras  \
+    ${BSPDIR}/sources/meta-pelux/meta-arp-extras    \
     ${BSPDIR}/sources/meta-intel                    \
     ${BSPDIR}/sources/meta-arp                      \
 "
--- a/meta-arp-extras/conf/layer.conf
+++ b/meta-arp-extras/conf/layer.conf
@@ -1,0 +1,12 @@
+BBPATH .= ":${LAYERDIR}"
+
+# We have a recipes directory, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "pelux-bsp-arp-layer"
+BBFILE_PATTERN_pelux-bsp-arp-layer := "^${LAYERDIR}/"
+
+BBFILE_PRIORITY_pelux-bsp-arp-layer = "10"
+
+LAYERSERIES_COMPAT_pelux-bsp-arp-layer = "sumo thud warrior"
+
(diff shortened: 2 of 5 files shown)


conf/variant/arp/bblayers.conf.sample
```diff
@@ -8,4 +8,5 @@ BBLAYERS += "\
     ${BSPDIR}/sources/meta-pelux/meta-intel-extras  \
```
sashko: could you please sort them out somehow in a logical way?

meta-arp-extras/conf/layer.conf
```diff
@@ -0,0 +1,12 @@
...
+
+# We have a recipes directory, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "pelux-bsp-intel-layer"
+BBFILE_PATTERN_pelux-bsp-intel-layer := "^${LAYERDIR}/"
+
+BBFILE_PRIORITY_pelux-bsp-intel-layer = "10"
+
+LAYERSERIES_COMPAT_pelux-bsp-intel-layer = "sumo thud warrior"
```
sashko: arp

meta-arp-extras/recipes-core/base-files/base-files/fstab
```diff
@@ -0,0 +1,5 @@
+/dev/root            /                    auto       defaults              1  1
```
sashko: fix indentations please

meta-arp-extras/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -0,0 +1,8 @@
+#
+#   Copyright (C) 2017 Pelagicore AB
```
sashko: 2019

meta-arp-extras/conf/layer.conf
```diff
@@ -0,0 +1,12 @@
+BBPATH .= ":${LAYERDIR}"
+
+# We have a recipes directory, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "pelux-bsp-intel-layer"
+BBFILE_PATTERN_pelux-bsp-intel-layer := "^${LAYERDIR}/"
+
+BBFILE_PRIORITY_pelux-bsp-intel-layer = "10"
```
sashko: arp

meta-arp-extras/conf/layer.conf
```diff
@@ -0,0 +1,12 @@
+BBPATH .= ":${LAYERDIR}"
+
+# We have a recipes directory, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "pelux-bsp-intel-layer"
```
sashko: arp

martin-ejdestig:
> ARP is supposed to use fstab from intel, but instead it was using the default one from poky. This commit contains the fstab that will be used on ARP.

Can this not be controlled by layer priorities? Does `bitbake -e <whatever it is that provides the fstab>` tell you why it does not? Compare what the difference is between an Intel build and an ARP one.

conf/variant/arp/bblayers.conf.sample
fhajredini: Done!

fhajredini:
> > ARP is supposed to use fstab from intel, but instead it was using the default one from poky. This commit contains the fstab that will be used on ARP.
> 
> Can this not be controlled by layer priorities? Does `bitbake -e <whatever it is that provides the fstab>` tell you why it does not? Compare what the difference is between an Intel build and an ARP one.

I went through the log and didnt notice a difference. I will go ahead and merge this as for now, this seems the only way to have this fixed.

PR #325, opened 2019-06-24
local.conf:create a bmap file

fhajredini:
Since our images have grown in size, dd-ing takes a long time. Bmap
tools allow us to take advantage of sparse files and reduce time needed
for flashing of an image.

### DIFF
--- a/conf/variant/arp-qtauto/local.conf.sample
+++ b/conf/variant/arp-qtauto/local.conf.sample
@@ -4,6 +4,8 @@ MACHINE = "arp"
 
 EFI_PROVIDER = "grub-efi"
 
+IMAGE_FSTYPES += "wic wic.bmap"
+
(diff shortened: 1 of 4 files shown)


sashko:
So... how exactly does it work? I got the idea of bmap-tools, but what do we have now - two types of images?

fhajredini:
There are going to be two files needed:

- *.rootfs.wic
- *.rootfs.wic.bmap

The bmap file takes 8KB while the wic file takes the usual 7.9GB. When flashing you use bmap-tools for which you use both of the files when flashing, the image and the bmap file.

Will extend the documentation on swf as well.

sashko:
Does it support wic images only?

You forgot space in the title

fhajredini:
I am currently testing a build on rpi by ading the image fs type. Not sure why do we not have a wic image for rpi as well. For now, it works on intel.

PR #327, opened 2019-06-26
Include mount parts of udev-extraconf in Poky

martin-ejdestig:
Will make it so pluggable mass storage devices are automatically mounted under /run/media/ when inserted.

Since PELUX has two root partitions for swupdate, one that is actually mounted at / and one that is not mounted at all, they need to be blacklisted from the script that is run when a mass storage device is
detected. If it is not, the rootfs that is not / will be mounted under /run/media/. mount.sh in udev-extraconf is extended with the functionality to use UUID:s or labels to blacklist devices so that this can be reliable done in PELUX (using the platform1 and platform2 labels).

Still some verification left to do. (E.g. not exactly sure why secondary / did not show up before the udev-extraconf patch was added on RPI.)

### DIFF
--- a/classes/core-image-pelux.bbclass
+++ b/classes/core-image-pelux.bbclass
@@ -19,6 +19,12 @@ IMAGE_INSTALL_append = "\
     ${@bb.utils.contains("COMBINED_FEATURES", "bluetooth", "packagegroup-tools-bluetooth", "", d)} \
 "
 
+# Include udev-extraconf for automatic mounting of pluggable mass storage
+# devices (other parts are stripped out).
+IMAGE_INSTALL_append = "\
+    udev-extraconf \
+"
+
 # GENIVI components
 IMAGE_INSTALL_append = "\
     dlt-daemon         \
--- a/recipes-core/udev-extraconf/udev-extraconf_%.bbappend
+++ b/recipes-core/udev-extraconf/udev-extraconf_%.bbappend
@@ -1,0 +1,16 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI +=  "\
+    file://0001-udev-extraconf-allow-labels-and-UUID-s-in-mount-blac.patch;striplevel=5 \
+    file://pelux-roots.blacklist \
+"
+
+do_install_append() {
+    install -m 0644 ${WORKDIR}/pelux-roots.blacklist ${D}${sysconfdir}/udev/mount.blacklist.d
+
+    # Only want mount parts of udev-extraconf.
+    rm ${D}${sysconfdir}/udev/rules.d/autonet.rules
+    rm ${D}${sysconfdir}/udev/rules.d/localextra.rules
+    rm ${D}${sysconfdir}/udev/scripts/network.sh
+}
+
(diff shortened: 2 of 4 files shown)


martin-ejdestig:
> Still some verification left to do. (E.g. not exactly sure why secondary / did not show up before the udev-extraconf patch was added on RPI.)

OK, turns out this is due to the secondary partition being created first when a software update is performed. Should set labels on RPI as well for blacklisting to work. Perhaps do it after #316 tough since it changes so that both root partitions will always be there even before an update.

classes/core-image-pelux.bbclass
```diff
@@ -19,6 +19,12 @@ IMAGE_INSTALL_append = "\
     ${@bb.utils.contains("COMBINED_FEATURES", "bluetooth", "packagegroup-tools-bluetooth", "", d)} \
 "
 
+# Include udev-extraconf for automatic mounting of pluggable mass storage
+# devices (other parts are stripped out).
+IMAGE_INSTALL_append = "\
```
sashko: We should stop adding IMAGE_INSTALL for every component and perhaps have a few generic sections.

martin-ejdestig: Hum, sure... but what do you suggest more concretely at this point? I can not see a way to group things together in a nice way at the moment. Perhaps just remove some comments and then just have 1 or 2... perhaps even 3... IMAGE_INSTALL? Most comments do not add anything... some things should not be here... should we have a "should be moved out of here" section?

sashko: I do not have a clear suggestion atm. Let's discuss this separately then. You can proceed with the current change.

recipes-core/udev-extraconf/udev-extraconf_%.bbappend
```diff
@@ -0,0 +1,16 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
```
sashko: this is not necessary since $thisdir/pn is a default search path.

martin-ejdestig: If I remove this line I get:

```
ERROR: udev-extraconf-1.1-r0 do_fetch: Fetcher failure for URL: 'file://0001-udev-extraconf-allow-labels-and-UUID-s-in-mount-blac.patch;striplevel=5'. Unable to fetch URL from any source.
```

sashko: strange: https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#var-FILESPATH

classes/core-image-pelux.bbclass
martin-ejdestig: I added an extra commit on top with a suggestion for how to reduce the current number of IMAGE_INSTALL_append.

sashko:
Make the second commit in a separate PR, please.

recipes-core/udev-extraconf/udev-extraconf_%.bbappend
martin-ejdestig: I do not see it saying there that it is not needed when you bbappend. https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#var-FILESEXTRAPATHS even says to do what we do in our bbappends. And they seem to do it in poky etc.

Anyway, if you still think it should work without and can figure out why it does not (I can not, I'll remove it. :)

sashko: no, I don't care. it was just a small tip.

PR #328, opened 2019-06-27
core-image-pelux: remove unnecessary IMAGE_FSTYPES

sashko:
We are not using sdcard anywhere and ext3 is already set in BSP layers,
e.g. meta-intel and meta-rapsberrypi.


martin-ejdestig:
Just to check, we will still get a *-sdcard image for RPI even after this change?

sashko:
> Just to check, we will still get a *-sdimg image for RPI even after this change?

yes, I verified.

PR #329, opened 2019-06-27
core-image-pelux-minimal-dev: remove tools-testapps

sashko:
tools-testapps packagegroup is comprised of tools we are not using.

LPT package alone takes about 20 minutes to build on a good machine
and occupies 250 MB of space on the filesystem.


sashko:
Package group in question: http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/packagegroups/packagegroup-core-tools-testapps.bb?h=master

PR #330, opened 2019-06-27
variant/local.conf.sample: Include bz2 type

fhajredini:

### DIFF
--- a/conf/variant/arp-qtauto/local.conf.sample
+++ b/conf/variant/arp-qtauto/local.conf.sample
@@ -4,7 +4,7 @@ MACHINE = "arp"
 
 EFI_PROVIDER = "grub-efi"
 
-IMAGE_FSTYPES += "wic wic.bmap"
+IMAGE_FSTYPES += "wic wic.bmap wic.bz2"
 
 PREFERRED_PROVIDER_iasl-native = "iasl-native"
 
--- a/conf/variant/arp/local.conf.sample
+++ b/conf/variant/arp/local.conf.sample
@@ -2,7 +2,7 @@ require conf/variant/common/local.conf
 
 MACHINE = "arp"
 
-IMAGE_FSTYPES += "wic wic.bmap"
+IMAGE_FSTYPES += "wic wic.bmap wic.bz2"
 
 EFI_PROVIDER = "grub-efi"
 
--- a/conf/variant/intel-qtauto/local.conf.sample
+++ b/conf/variant/intel-qtauto/local.conf.sample
@@ -2,7 +2,7 @@ require conf/variant/common/local.conf
 
 MACHINE = "intel-corei7-64"
 
-IMAGE_FSTYPES += "wic wic.bmap"
+IMAGE_FSTYPES += "wic wic.bmap wic.bz2"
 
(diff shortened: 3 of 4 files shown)


sashko:
what for?

fhajredini:
> what for?

Updated the commit message

sashko:
it will be a compressed image (~1GB) and it will be faster to download
    it supports compressed images for faster flashing to disk.
waot, what? so will we get, two images: wic and wic.bz2? how can one flash bz2 one without decompressing it?

martin-ejdestig:
> how can one flash bz2 one without decompressing it?

bmaptool takes care of decompressing it: https://github.com/intel/bmap-tools/blob/master/docs/README#L236

(Presumably, I have not tried it myself yet. :)

sashko:
> > how can one flash bz2 one without decompressing it?
> 
> bmaptool takes care of decompressing it: https://github.com/intel/bmap-tools/blob/master/docs/README#L236
> 
> (Presumably, I have not tried it myself yet. :)

okay, great. @fhajredini can you confirm? also, double check whether we need `wic.bmap` after this change or `wic.bmap.bz2` would be enough. don't forget about updating docs as well.

fhajredini:
> > > how can one flash bz2 one without decompressing it?
> > 
> > 
> > bmaptool takes care of decompressing it: https://github.com/intel/bmap-tools/blob/master/docs/README#L236
> > (Presumably, I have not tried it myself yet. :)
> 
> okay, great. @fhajredini can you confirm? also, double check whether we need `wic.bmap` after this change or `wic.bmap.bz2` would be enough. don't forget about updating docs as well.

We still need wic.bmap

sashko:
local.conf.sample: add wic.bmap.bz2 to IMAGE_FSTYPES / generate wic.bmap.bz2 IMAGE_FSTYPE for Intel and ARP

fhajredini:
Just checked again, it can be flashed without bmap file too, but that wouldnt be so smart as it increases the time of the flashing a lot. I think that its the bmap file which takes use of the sparse bytes..

sashko:
> Just checked again, it can be flashed without bmap file too, but that wouldnt be so smart as it increases the time of the flashing a lot. I think that its the bmap file which takes use of the sparse bytes..

I am not sure I understood you: isn't wic.bmap.bz just a compressed version of wic.bmap?

fhajredini:
> > Just checked again, it can be flashed without bmap file too, but that wouldnt be so smart as it increases the time of the flashing a lot. I think that its the bmap file which takes use of the sparse bytes..
> 
> I am not sure I understood you: isn't wic.bmap.bz just a compressed version of wic.bmap?

No, wic.bmap is a file of 8KB. As far as i understand, its just a file which maps the image for easier flashing.. now the actual image is still required, and that can be a raw image(.wic) or a compressed image(.wic.bz2). Bmaptools can flash a compressed image on the fly. It can compress it with, or without the bmap file, but without it, it will take a longer time.

sashko:
title is still ambiguous

fhajredini:
bmap has already been included on the previous commits. This is just having a compressed image as an artifact.

PR #336, opened 2019-07-03
swupdate: add runtime dependency on blktool

sashko:
SWUpdate needs blktool to function.


martin-ejdestig:
Hum, I do not see how blktool is used by swupdate.

PR #338, opened 2019-07-08
qtwayland: fix config for Raspberry Pi

sashko:
We use eglfs as a platform plugin to run Neptune 3 UI. Moreover, it
breaks build of QtWayland 5.13 for a reason that is currently not
understood.


martin-ejdestig:
> We use eglfs as a platform plugin

I do not understand. EGLFS has a Rasberry Pi backend. We use EGLFS for RPI on master? wayland-bcrm is just for some wayland specific RPI integration?

martin-ejdestig:
What does wayland-bcrm in qtwayland actually do? Would be bad if we disable it and don't get any acceleration due to e.g. wrong formats being used.

PR #344, opened 2019-07-19
udev-extraconf: do not mount swap partitions

sashko:
Skip mounting swap partitions as they have special filesystem that
cannot and should not be mounted.


martin-ejdestig:
I guess swap is not that common in cases where this script is used. :)

PR #347, opened 2019-07-22
Patches to make persistance-administrator daemon run

martin-ejdestig:
This commit only makes it so it runs without immediately exiting when trying to start the pas-daemon service. See commit messages in patch files and upstream pull requests for details.

Note: I think we should wait with pulling this into master before we have verified the full pcl vertical. Having this PR makes it easier for others to pick up the persistance-administrator fixes at least.


sashko:
why not use `persistance-administrator` tag

martin-ejdestig:
We usually tag per component/recipe... this commit modifies 3 of them... ? (One of the downsides with tags. Tags are optional anyway.)

martin-ejdestig:
> Tags are optional anyway.

That said, if most commits in a repo use tags and there is a tag that makes sense, it makes sense to have one, I guess.

PR #348, opened 2019-07-23
cleanup /etc/fstab handling for Intel and ARP

martin-ejdestig:
Work in progress. Will hold off with this until after 4.0 release.


martin-ejdestig:
Should perhaps consider removing the hardcoded UUID:s completely. As far as I can tell they are only used to specify PARTUUID in bootloader for root. But labels should work there as well, I think.

conf/variant/arp-qtauto/bblayers.conf.sample
```diff
@@ -7,7 +7,6 @@ require conf/variant/common/bblayers-qtauto.conf
 
 BBLAYERS += "\
     ${BSPDIR}/sources/meta-pelux/meta-intel-extras  \
-    ${BSPDIR}/sources/meta-pelux/meta-arp-extras    \
```
sashko: why?

martin-ejdestig: It was only added for the /etc/fstab workaround.

martin-ejdestig:
4.0 has been released and no longer wip, ready for review and merge.

PR #350, opened 2019-07-25
connman: add connman-client to rdepends

sashko:
We need connman-client to work with connman. Add it to its runtime
dependencies, so it is always deployed to images together with connman.

### DIFF
--- a/recipes-connectivity/connman/connman_%.bbappend
+++ b/recipes-connectivity/connman/connman_%.bbappend
@@ -1,3 +1,5 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 SRC_URI += "file://disable_connman_dns_proxy.patch"
+
+RDEPENDS_${PN} += " connman-client"


recipes-connectivity/connman/connman_%.bbappend
```diff
@@ -1,3 +1,5 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 SRC_URI += "file://disable_connman_dns_proxy.patch"
+
+RDEPENDS += " connman-client"
```
martin-ejdestig: Don't you need `_${PN}` here?

And will this not cause a circular dependency since the ConnMan recipe has this: `RDEPENDS_${PN}-client ="${PN}"`

martin-ejdestig: Or does this actually work as is? Tested?

sashko: you do

sashko: circular dependency is handled fine

martin-ejdestig: Hum, yeah. I guess it does not matter that it is circular since it is run-time. OK, if this works it seems like a better solution than Pelagicore/meta-bistro#254 .

PR #353, opened 2019-07-31
rpi-config: do not set GPU_MEM

sashko:
GPU_MEM is set to 256 Mb in meta-boot2qt now and is sufficient to run
Neptune 3 UI, threfore it makes sence to save precious RAM.


martin-ejdestig:
I have not checked when/why this was changed in meta-boot2qt.

PR #354, opened 2019-08-07
Revert "Cleanup /etc/fstab handling for Intel and ARP"

TSAnsari:
This reverts commit 1acf4025a35a630a572d31938ada529951d6c153.

Even though automatically generating fstab from the wks file
is a nice solution, but it does not work well with the swupdate
feature. The reason being that the swupdate swu file does not
contain the fstab file that is generated for wic image. And there
seems to be no way of updating the swu file to contain the fstab
generated by wic script. So the swu file uses the base fstab from
the rpi layer. This results in the partitions mounted incorrectly
when the partition is updated using swupdate. So here we revert
to the solution that uses hard coded fstab files instead.

### DIFF
--- a/conf/variant/arp-qtauto/bblayers.conf.sample
+++ b/conf/variant/arp-qtauto/bblayers.conf.sample
@@ -7,6 +7,7 @@ require conf/variant/common/bblayers-qtauto.conf
 
 BBLAYERS += "\
     ${BSPDIR}/sources/meta-pelux/meta-intel-extras  \
+    ${BSPDIR}/sources/meta-pelux/meta-arp-extras    \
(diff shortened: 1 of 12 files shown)


sashko:
what is going to be the solution?

TSAnsari:
> what is going to be the solution?

The solution is what this commit has reverted to.

TSAnsari:
> Too bad that there does not seem like there is a way to have `wic` generate the fstab for *.swu. :(
> 
> Only a minor nit in commit message:
> 
> > So the swu file uses the base fstab from the rpi layer.
> 
> I think the default is taken from `base-files` in `poky` for Intel and ARP.

Yes. Sorry, I don't know why I was thinking about rpi. Will fix it

PR #355, opened 2019-08-08
fstab: mount boot and data partition

TSAnsari:
RPI image was changed to wic format as well as
a new data partition was added to the image in
the following commits,

26079f1d6af92298494754f8d146cdcdc3edbf69
0d70eaf4fb638504ed9ed8bb5b09e4efae4fda60

fstab was not updated to mount the partitions
added to the wks file. This commit adds these
partitions to fstab.

WIC_CREATE_EXTRA_ARGS = "--no-fstab-update" is
also added to the local.conf for rpi. This stops
wic script to generated fstab file from wks file.
This is required to support swupdate swu image to
use the hardcoded fstab file instead.


scripts/lib/wic/canned-wks/raspberrypi.wks
```diff
@@ -2,7 +2,7 @@
 # long-description: Creates a partitioned SD card image for use with
 # Raspberry Pi. Boot files are located in the first vfat partition.
 
-part /boot --source bootimg-partition --ondisk mmcblk0 --fstype=vfat --label boot --active --align 4096 --size 20
-part / --source rootfs --ondisk mmcblk0 --fstype=ext4 --label platform1 --align 4096 --uuid b488d900-9cc0-11e9-8833-cf4cb425a04d
-part / --source rootfs --ondisk mmcblk0 --fstype=ext4 --label platform2 --align 4096 --uuid a2d8ad56-9c9e-11e9-afd8-07ffb5f9d79e
-part /data --ondisk mmcblk0 --size 1024 --fstype=ext4 --label data --align 4096 --uuid 79664834-9c9e-11e9-9603-a744b9801557
+part /boot --source bootimg-partition --ondisk mmcblk0 --fstype=vfat --label boot --active --align 4096
```
martin-ejdestig: Is removing "--size 20" intentional?

PR #357, opened 2019-08-08
systemd: stop low-level messages on console

sashko:
Lower kernel.printk values to the ones Ubuntu has to get rid of
unnecessary low-level messages.

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -1,3 +1,7 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += "file://99-printk.conf"
+
 do_install_append() {
     # Make sure resolv.conf is using systemd-resolved
     ln -s /run/systemd/resolve/resolv.conf ${D}${sysconfdir}/resolv.conf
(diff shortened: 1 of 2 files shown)


sashko:
Discussed in https://github.com/Pelagicore/meta-bistro/pull/256

sashko:
@martin-ejdestig ping

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -11,5 +15,11 @@ do_install_append() {
 
 FILES_${PN} += "/etc/resolv.conf"
 FILES_${PN} += "${sysconfdir}/systemd/system/getty.target.wants"
+FILES_${PN} += "/lib/sysctl.d/99-printk.conf"
```
martin-ejdestig: ${exec_prefix}/lib/sysctl.d ? Will expand to /usr/lib/sysctl.d, I think. What the systemd recipe uses:

http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/systemd/systemd_239.bb?h=thud#n540

Actually, I do not see a /lib/sysctl.d directory on target. Have you tested that this actually works? Regardless, using variables should give you correct location...?

sashko: so, where should this be deployed, `/lib` or `/usr/lib` ?

PR #360, opened 2019-08-13
core-image-pelux: add iproute2

sashko:
iproute2 is required to setup CAN devices on Linux.


sashko:
> Wondering a bit how, only for some specific device or all, etc. ... but... approved.

https://developer.ridgerun.com/wiki/index.php/How_to_configure_and_use_CAN_bus#Enable_socketcan_and_ip_tools

It's just that `ip` tool from `busybox` won't do the job.

PR #361, opened 2019-08-19
systemd-resource-control: add recipe

sashko:
Add a bunch of slice units for the future use.

### DIFF
--- a/recipes-config/systemd-resource-control/systemd-resource-control.bb
+++ b/recipes-config/systemd-resource-control/systemd-resource-control.bb
@@ -1,0 +1,28 @@
+SUMMARY = "A set of systemd slice units for PELUX"
+DESCRIPTION = "This package supplies a set of generic systemd slice units that \
+can be used by different services in PELUX."
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=3e927855e47b22eecd81954b0f545a0f"
+
+SRC_URI += " \
+    file://LICENSE \
+    file://connectivity.slice \
+    file://ui.slice \
+"
+
+S = "${WORKDIR}"
+
+inherit allarch
+
+do_install_append() {
+    install -d ${D}${systemd_unitdir}/system/
+
+    install -m 0644 ${WORKDIR}/connectivity.slice ${D}${systemd_unitdir}/system/
+    install -m 0644 ${WORKDIR}/ui.slice ${D}${systemd_unitdir}/system/
+}
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/slices/:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/systemd-resource-control/:"
+FILES_${PN} += " \
+    ${systemd_unitdir}/system/ \
+"
--- a/recipes-config/systemd-resource-control/slices/connectivity.slice
+++ b/recipes-config/systemd-resource-control/slices/connectivity.slice
@@ -1,0 +1,3 @@
+[Unit]
+Description=Connectivity Slice
+Before=slices.target
--- a/recipes-config/systemd-resource-control/slices/ui.slice
+++ b/recipes-config/systemd-resource-control/slices/ui.slice
@@ -1,0 +1,3 @@
+[Unit]
+Description=UI Slice
+Before=slices.target
--- a/recipes-config/systemd-resource-control/systemd-resource-control/LICENSE
+++ b/recipes-config/systemd-resource-control/systemd-resource-control/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2019 Luxoft Sweden AB
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.


recipes-config/systemd-resource-managament/systemd-resource-managament.bb
```diff
@@ -0,0 +1,84 @@
...
+
+do_install_append() {
+    install -d ${D}${systemd_unitdir}/system/
+
+    # Slices
+    install -d ${D}${systemd_unitdir}/system/audio.slice.d/
+    install -m 0644 ${WORKDIR}/audio.slice ${D}${systemd_unitdir}/system/audio.conf
+
+    install -d ${D}${systemd_unitdir}/system/connectivity.slice.d/
+    install -m 0644 ${WORKDIR}/connectivity.slice ${D}${systemd_unitdir}/system/connectivity.conf
```
JEderonn: typo, destination file name prefix shall be slice, not conf.

But, the question is however if we shall create a slice configuration file (slice.conf) and put it as an override into the <slice name>.slice.d/ folder instead of creating the slice unit file in the /system folder.

See other comment

recipes-config/systemd-resource-managament/slices/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2019 Luxoft Sweden AB
```
JEderonn: Is the intention to have the LICENSE file in the slices folder?

recipes-config/systemd-resource-managament/slices/uiapps.slice
```diff
@@ -0,0 +1,3 @@
+[Unit]
```
JEderonn: Rename the slice from uiapps to ui-apps, indicating that its parent is ui.

sashko:
> I'm wondering if we only shall add the connectivity slice and assign it to the connectivity-manager service for now. Then, we can add more slices and service configurations in another commit.
> 
> So, consider remove all slice unit files but connectivity.slice. And assign connman.service to connectivity.slice as you do for the connectivity-manager.

This is WIP. I've added one as an example.

recipes-config/systemd-resource-managament/systemd-resource-managament.bb
JEderonn: I'm thinking of 2 scenarios:
1. Create a new slice, like connectivity. In this case, one may create a new slice unit file, like connectivity.slice, and all slice configurations are put here and finally installed into /lib/systemd/system/<slice name>.slice.
2. Adapt an existing slice, like system.slice. In this case, one may create a override by adding slice.system.conf to the layer which is then installed into the /lib/systemd/system/system.slice.d/slice.conf.

For the services, it's all about override becase it's assumed that the service exists. So in this case, the configuration file installed into /lib/systemd/system/<service name>.service.d/service.conf.

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -0,0 +1,86 @@
...
+    install -m 0644 ${WORKDIR}/slice.ui.conf ${D}${systemd_unitdir}/system/ui.slice
+
+    install -d ${D}${systemd_unitdir}/system/vehiclefunctions.slice.d/
+    install -m 0644 ${WORKDIR}/slice.vehiclefunctions.conf ${D}${systemd_unitdir}/system/vehiclefunctions.slice
+
+    install -d ${D}${systemd_unitdir}/system/video.slice.d/
+    install -m 0644 ${WORKDIR}/video.slice ${D}${systemd_unitdir}/system/video.conf
+
+    # Service appends
+    install -m 0644 ${WORKDIR}/service.connectivity-manager.conf ${D}${systemd_unitdir}/system/connectivity.slice.d/
```
JEderonn: typo, destination folder shall be named connectivity-manager.service.d/service.conf.

recipes-config/systemd-resource-control/slices/slice.connectivity.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
```
JEderonn: Rename slice.connectivity.conf to connectivity.slice if you agree that we shall not use the override feature for slice unit configuration (only for service unit configuration / adaptations)

recipes-config/systemd-resource-control/services/service.connectivity-manager.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
```
JEderonn: Typo, shall be in [Service] block.

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -0,0 +1,86 @@
...
+}
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/services:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/slices:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/systemd-resource-control/:"
+FILES_${PN} += " \
+    ${systemd_unitdir}/system/ \
+    ${systemd_unitdir}/system/audio.slice.d/ \
```
JEderonn: Add the missing connectivity-manager.service.d folder.

sashko: but it's not a slice

JEderonn: You are correct, but the folder is created by this recipe, i.e. ${D}${systemd_unitdir}/system/connectivity-manager.service.d/service.conf. Shouldn't it be part of FILES_${PN} then?

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -0,0 +1,87 @@
...
+inherit allarch
+
+do_install_append() {
+    install -d ${D}${systemd_unitdir}/system/
+
+    # Slices
+    install -d ${D}${systemd_unitdir}/system/audio.slice.d/
+    install -m 0644 ${WORKDIR}/slice.audio.conf ${D}${systemd_unitdir}/system/audio.slice
+
+    install -d ${D}${systemd_unitdir}/system/connectivity.slice.d/
```
JEderonn: delete row since we don't have any slice overrides for connectivity at the moment.

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -0,0 +1,87 @@
...
+S = "${WORKDIR}"
+
+inherit allarch
+
+do_install_append() {
+    install -d ${D}${systemd_unitdir}/system/
+
+    # Slices
+    install -d ${D}${systemd_unitdir}/system/audio.slice.d/
+    install -m 0644 ${WORKDIR}/slice.audio.conf ${D}${systemd_unitdir}/system/audio.slice
```
JEderonn: typo, source slice.audio.conf shall be audio.slice.

recipes-config/systemd-resource-control/systemd-resource-control.bb
JEderonn: Furthermore, the slice override folders are not installed because we don't install anything in them, so I assume they shall be removed from this variable?

JEderonn:
Could you please add a README file with a short description of its purpose and usage.

Furthermore, consider to rename the package to systemd-resource-control.

sashko:
> Could you please add a README file with a short description of its purpose and usage.

Added `SUMMARY` and `DESCRIPTION` fields to the recipe.

> Furthermore, consider to rename the package to systemd-resource-control.

Done.

PR #364, opened 2019-08-20
pelux.conf: cleanup

sashko:
Removed DISTRO_FEATURES that are already part of DISTRO_FEATURES_DEFAULT
variable in oe-core and cleanedup syntax.

### DIFF
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -17,25 +17,15 @@ TARGET_VENDOR = "-pelux"
 
 # Add distro features
 DISTRO_FEATURES_append = " \
-                          alsa         \
-                          argp         \
-                          bluetooth    \
-                          bluez5       \
-                          largefile    \
-                          multiarch    \
-                          opengl       \
-                          pci          \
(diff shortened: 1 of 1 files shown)


sashko:
For the reference: https://github.com/openembedded/openembedded-core/blob/master/meta/conf/distro/include/default-distrovars.inc#L13

conf/distro/pelux.conf
```diff
@@ -17,25 +17,15 @@ TARGET_VENDOR = "-pelux"
...
 # Add distro features
 DISTRO_FEATURES_append = " \
-                          alsa         \
-                          argp         \
-                          bluetooth    \
-                          bluez5       \
-                          largefile    \
-                          multiarch    \
-                          opengl       \
-                          pci          \
```
martin-ejdestig: pci intentionally removed?

sashko: It is part of the default configuration.

martin-ejdestig: Ah, I did not see it since the list is not sorted.

PR #365, opened 2019-08-21
grub: remove disabling execlist support for i915

sashko:
This has been added in @e5428651123e47d3c, since the Linux kernel
driver supplied wth Yocto "sumo" 2.5 had an issue that caused it to
crash.

The issue has resolved fixed in newer versions of the driver, so get
rid of this workaround.

### DIFF
--- a/scripts/lib/wic/canned-wks/cfg
+++ b/scripts/lib/wic/canned-wks/cfg
@@ -8,9 +8,9 @@ save_env default
 
 # Boot entries
 menuentry 'bootA'{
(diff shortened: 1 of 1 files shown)


sashko:
> Approving... but Arch Wiki referred to in [e542865](https://github.com/Pelagicore/meta-pelux/commit/e5428651123e47d3c42c51c0b1dd8adf2ebcc427) says "This is known to be broken to at least kernel 4.0.5.". We are not on >= 4.0.5 but fix has been cherry picked?

We've got `4.14.110` for Intel.

PR #367, opened 2019-08-22
layer.conf: add bc to HOSTTOOLS

sashko:
bc is required to build kernel and kernel modules with the SDK,
otherwise the following error will occur:

  /bin/sh: 1: bc: not found


sashko:
bc has to be part of the Docker image for a build to succeed: https://github.com/Pelagicore/pelux-manifests/pull/421

PR #369, opened 2019-08-23
Quiet boot

sashko:


scripts/lib/wic/canned-wks/cfg
```diff
@@ -1,6 +1,6 @@
 serial --unit=0 --speed=115200 --word=8 --parity=no --stop=1
 
-timeout=5
+timeout=1
```
martin-ejdestig: I actually do not think this is that interesting. Not something to take into account when measuring boot time.

If we want to reduce it for demonstration purposes we should probably just remove the menu altogether. If I remember correctly one can get the menu to show during boot by pressing ESC.

martin-ejdestig:
> grup: use quiet boot 

Nit: grup -> grub

Only for RPI? Do the same for RPI or wait with it for now? Would also be interesting to know approximately how much time is saved.

scripts/lib/wic/canned-wks/cfg
sashko: Yes, I completely agree. This is not a practical optimization of any kind. It is theoretically possible to gain access to the boot menu even if `timeout=0`, but I did not manage to do so :( I can drop this change if someone insists.

sashko:
> Only for RPI? Do the same for RPI or wait with it for now?

Could you please elaborate that?

> Would also be interesting to know approximately how much time is saved.

If there are no errors printed on the console (#370 fixes one), ~1500 ms.

scripts/lib/wic/canned-wks/cfg
martin-ejdestig: I am not against it per se but think we should remove it completely if we want a "smooth experience (tm)". Pressing ESC to get it to show does not work? Maybe do this change now and add a sub-task to the boot time task for investigating later if you can not get it to show?

martin-ejdestig:
> > Only for RPI? Do the same for RPI or wait with it for now?
> 
> Could you please elaborate that?

This is only for Intel/ARP but we want to add `quiet` for RPI as well?

> > Would also be interesting to know approximately how much time is saved.
> 
> If there are no errors printed on the console (#370 fixes one), ~1500 ms.

Nice.

martin-ejdestig:
Still should change "grup:" to "grub:" in second commit.

And... I guess you want to punt adding quiet for RPI... ?

And you have not answered if "ESC" worked or not if the menu is completely hidden. I am fine with punting this if we track it somewhere (assuming this is behavior we want... I think it is... hide grub completely, show splash as soon as possible... and make it so the grub menu can be brought up if really wanted).

PR #370, opened 2019-08-23
pelux.conf: add pam to DISTRO_FEATURES

sashko:
systemd needs PAM support for user sessions to work correctly,
otherwise XDG_RUNTIME_DIR won't be created on boot and Wayland
will complain[1]:

  [FAILED] Failed to start User Manager for UID 0.

pam packageconfig has been enabled in meta-boot2qt (@ded06350f2dd23b),
and will later be merged to thud and/or warrior branche, but
prerequisite for it is pam DISTRO_FEATURE in Yocto.

[1] https://bugreports.qt.io/browse/QTBUG-75543

### DIFF
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -20,6 +20,7 @@ DISTRO_FEATURES_append = " \
     bluez5 \
     multiarch \
     opengl \
+    pam \
     process-containment \
     ptest \
     virtualization \


martin-ejdestig:
Is this a problem we actually have when running Neptune as a service or something that is only relevant if we use proper login sessions? That is, have you seen this error and this fixes it?

Looking at the systemd code it looks like XDG_RUNTIME_DIR is only ever set by the pam_systemd module when a session is opened: https://github.com/systemd/systemd/blob/58c1a34437f2560c967318ad75d0615bd33b90b1/src/login/pam_systemd.c#L458

So by just looking at the systemd code I do not see how this would fix anything, but most likely I am missing something.

sashko:
> Is this a problem we actually have when running Neptune..? That is, have you seen this error and this fixes it?

Yes, and yes.

> Looking at the systemd code it looks like XDG_RUNTIME_DIR is only ever set by the pam_systemd module when a session is opened: https://github.com/systemd/systemd/blob/58c1a34437f2560c967318ad75d0615bd33b90b1/src/login/pam_systemd.c#L458
> 
> So by just looking at the systemd code I do not see how this would fix anything, but most likely I am missing something.

This will add `pam` packageconfig to systemd: https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-core/systemd/systemd_239.bb?h=thud#n84

PR #373, opened 2019-08-26
thermald: disable systemd service

sashko:
thermald comes with BSP package in meta-intel and although its zero
configuration mode is normally sufficient for most systems, there is
currently no need in any thermal control.

Also, due to the lack of configuration supplied, it produces a bunch
of errors, flooding the systemd log:

  sensor id 6 : No temp sysfs for reading raw temp
  sensor id 6 : No temp sysfs for reading raw temp
  sensor id 6 : No temp sysfs for reading raw temp
  I/O warning :
    failed to load external entity "/etc/thermald/thermal-conf.xml"
  error: could not parse file /etc/thermald/thermal-conf.xml
  sysfs open failed
  I/O warning :
    failed to load external entity "/etc/thermald/thermal-conf.xml"
  error: could not parse file /etc/thermald/thermal-conf.xml
  I/O warning :
    failed to load external entity "/etc/thermald/thermal-conf.xml"
  error: could not parse file /etc/thermald/thermal-conf.xml

According to systemd-analyze blame, it takes ~2.5 seconds to initialize
it on Intel NUC:

  2.366s thermald.service

### DIFF
--- a/meta-intel-extras/recipes-bsp/thermald/thermald_%.bbappend
+++ b/meta-intel-extras/recipes-bsp/thermald/thermald_%.bbappend
@@ -1,0 +1,1 @@
+SYSTEMD_AUTO_ENABLE = "disable"


martin-ejdestig:
Any reason to keep building it?

sashko:
> Any reason to keep building it?

Not sure it's easy to disable: https://git.yoctoproject.org/cgit/cgit.cgi/meta-intel/tree/conf/machine/include/meta-intel.inc?h=thud#n17

martin-ejdestig:
A _remove will not work on MACHINE_ESSENTIAL_EXTRA_RDEPENDS ?

Looks like it was moved in https://git.yoctoproject.org/cgit/cgit.cgi/meta-intel/commit/?id=3b8d46f617277946391cc2ce62c953a4c5c9fc16 with no reason as to why it is always needed instead of just recommended. The comment above makes it sound like related to microcode loading but it looks like an oversight (comment not updated or thermald should have been added on a separate line to avoid confusion...).

JEderonn:
Even though thermald is not a vital package, it's a recommended package.
So why not adding a valid configuration for a NUC target? That would remove the error logs saying that it can't find the configuration file.
There are no dependencies to/from other services on thermald, so it shouldn't affect the startup performance.

sashko:
> Even though thermald is not a vital package, it's a recommended package.
> So why not adding a valid configuration for a NUC target? That would remove the error logs saying that it can't find the configuration file.

Sure, we could do this in a separate story. It might be a good idea to have it for ARP.

> There are no dependencies to/from other services on thermald, so it shouldn't affect the startup performance.

Agree.

martin-ejdestig:
> > There are no dependencies to/from other services on thermald, so it shouldn't affect the startup performance.
> 
> Agree.

Now I am confused. The service is always started, not configured to be D-Bus activated. Anyway, I do not know one way or another if it is worth disabling. Just wanted to point out that it is always started so can affect total time it takes to boot.

sashko:
> > > There are no dependencies to/from other services on thermald, so it shouldn't affect the startup performance.
> > 
> > 
> > Agree.
> 
> Now I am confused. The service is always started, not configured to be D-Bus activated. Anyway, I do not know one way or another if it is worth disabling. Just wanted to point out that it is always started so can affect total time it takes to boot.

Sorry, I must have answered the wrong question. That's why I want to disable it - it is started even if it is not configured and not needed.

JEderonn:
> Just wanted to point out that it is always started so can affect total time it takes to boot.

Sure, I agree. Let's keep it disabled.

martin-ejdestig:
Would be nice to know what the consequences are if it is disabled and why Intel think it is so important that they always include it (have never seen it running on a "desktop system").

PR #374, opened 2019-08-27
linux-intel: disable CONFIG_BTRFS_FS

sashko:
During boot, kernel spends around 200ms executing raid6_select_algo()
function.

RAID6 is selected by CONFIG_BTRFS_FS, neither of which we do not need.


sashko:
This also reduces image size by ~600Kb.

sashko:
> Do you intend to make the same change for other HW platforms like RaspberryPi too?

I do not see it on bootgraph on rpi3.

It seems to be set to 'm': https://github.com/raspberrypi/linux/blob/067fe10f2379864196d873ae0f4bd849c7b3950a/arch/arm64/configs/bcmrpi3_defconfig#L1125

PR #375, opened 2019-08-28
grub-efi: strip unnecessary modules

sashko:
grub is by default built with the following modules: boot, linux, ext2,
fat, serial, part_msdos, part_gpt, normal, efi_gop, iso9660, configfile,
search, loadenv, test.

Many of those modules are not needed neither for PELUX, neither for a
typical embedded Linux/IVI system.

Build grub-efi with necessary modules only to reduce binary size and
threfore boot time.


sashko:
> Approving... but... is this worth it? How much time is saved?

Honestly, it is a very minor improvement, but easy to do.

PR #376, opened 2019-08-29
mopidy: run service as mopidy user

sashko:
Create mopidy user and group and run systemd service as with those
permissions.


martin-ejdestig:
Unsure what route we want to go more generally with services. Perhaps something that should be discussed/spec:ed out a bit more? @JEderonn ?

PR #377, opened 2019-09-02
systemd-resource-control: add few services to connectivity slice

sashko:


recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -27,6 +28,7 @@ do_install_append() {
 
     # Install overrides for services
     install -d ${D}${systemd_unitdir}/system/neptune.service.d/
+    install -m 0644 ${WORKDIR}/service.bluetooth.conf ${D}${systemd_unitdir}/system/connman.service.d/
```
martin-ejdestig: Destination directory looks wrong.

If you do not know about it already you can verify overrides with `systemctl cat`. Comments with file the lines comes from will be added in output.

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -29,6 +30,7 @@ do_install_append() {
     # Install overrides for services
     install -d ${D}${systemd_unitdir}/system/neptune.service.d/
     install -m 0644 ${WORKDIR}/service.bluetooth.conf ${D}${systemd_unitdir}/system/connman.service.d/
+    install -m 0644 ${WORKDIR}/service.brcm43438.conf ${D}${systemd_unitdir}/system/connman.service.d/
```
martin-ejdestig: brcm43438 is only relevant for RPI?

JEderonn: Yes, let's wait with BSP specific services for now.
Furthermore, bluetooth.conf shall be installed into bluetooth.service.d/ not connman.service.d.

recipes-config/systemd-resource-control/services/service.brcm43438.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
```
JEderonn: Let's wait with BSP specific services until we have a solution to handle variants.

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -8,7 +8,12 @@ SRC_URI += " \
     file://LICENSE \
     file://connectivity.slice \
     file://ui.slice \
+    file://service.bluetooth.conf \
+    file://service.brcm43438.conf \
```
JEderonn: remove

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -24,13 +29,23 @@ do_install_append() {
...
+    install -m 0644 ${WORKDIR}/service.wpa_supplicant.conf ${D}${systemd_unitdir}/system/wpa_supplicant.service.d/
 }
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/services/:"
 FILESEXTRAPATHS_prepend := "${THISDIR}/slices/:"
 FILESEXTRAPATHS_prepend := "${THISDIR}/systemd-resource-control/:"
 FILES_${PN} += " \
     ${systemd_unitdir}/system/ \
+    ${systemd_unitdir}/system/bluetooth.service.d/ \
+    ${systemd_unitdir}/system/brcm43438.service.d/ \
```
JEderonn: remove brcm4348.service.d

recipes-config/systemd-resource-control/systemd-resource-control.bb
```diff
@@ -8,7 +8,12 @@ SRC_URI += " \
     file://LICENSE \
     file://connectivity.slice \
     file://ui.slice \
+    file://service.bluetooth.conf \
```
JEderonn: Does it make sense to only add the file if DISTRO_FEATURES contains bluetooth, like:

`${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'file://service.bluetooth.conf', '', d)} 
`

martin-ejdestig: A bit of a similar question... is there a way to say "if package installed, include this override". 

As far as I can tell there is no variable one can query with bb.utils.contains that contains all packages that will be installed, or is there?

E.g. if `connectivity-manager` is removed, this recipe will have to be updated as well. Easy to miss. But if it is complicated to install overrides conditionally in this package, it is perhaps better to live with potentially missing removing them if certain packages are not installed...?

JEderonn: > As far as I can tell there is no variable one can query with bb.utils.contains that contains all packages that will be installed, or is there?

Not that I'm aware of. I think you need to check this as part of a ROOTFS_POSTPROCESS_COMMAND.

```
my_postprocess_function() {
   echo "hello" > ${IMAGE_ROOTFS}/hello.txt
}
ROOTFS_POSTPROCESS_COMMAND += "my_postprocess_function; "
```

JEderonn: > E.g. if `connectivity-manager` is removed, this recipe will have to be updated as well. Easy to miss. But if it is complicated to install overrides conditionally in this package, it is perhaps better to live with potentially missing removing them if certain packages are not installed...?

Yes, so perhaps it's better to have a rootfs postprocess command which can clean overrides

```
for all folders matching xxx.d in ${IMAGE_ROOTFS}/lib/systemd/system
  IF xxx.d EXISTS BUT NOT xxx THEN rm -f xxx.d
```

recipes-config/systemd-resource-control/systemd-resource-control.bb
martin-ejdestig: Unresolving. `service.bluetooth.conf` still overrides `connman.service` instead of `bluetooth.serivce` it looks like to me. Or what am I missing?

JEderonn: `service.bluetooth.conf `shall be installed into folder `bluetooth.service.d/`, not `connman.service.d/`

JEderonn: I'm wondering if it would be better that the script is able to search for all files in the folder `services `which matches the syntax `service.*.conf `and installs them into `*.service.d/`.

recipes-config/systemd-resource-control/systemd-resource-control.bb
martin-ejdestig: Or perhaps just warn (or maybe even fail) if an override is detected for a service that is not installed ("X has been removed from dist => must update this recipe")?

sashko: Keep in mind having an append/override for nonexistent service is not going to break anything.

JEderonn: > Or perhaps just warn (or maybe even fail) if an override is detected for a service that is not installed ("X has been removed from dist => must update this recipe")?

Yes, a report would good.

JEderonn: > Keep in mind having an append/override for nonexistent service is not going to break anything.

You are right, and a report, part of the rootfs postprocess, would highlight these inconsistency issues.

PR #379, opened 2019-09-03
systemd: remove networkd 

sashko:


sashko:
this is how package config looks with those two patches:

`PACKAGECONFIG="     ldconfig pam     rfkill          acl     backlight     binfmt     firstboot     gshadow     hibernate     hostnamed     ima     kmod     localed     logind     machined     myhostname          nss     polkit     quotacheck     randomseed     resolved     smack     sysusers     timedated     timesyncd     utmp     vconsole     xz   "`

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -20,8 +20,6 @@ FILES_${PN} += " \
     ${libdir}/sysctl.d/99-printk.conf \
 "
 
-PACKAGECONFIG_append = " kmod "
```
martin-ejdestig: Do you know why this was added here in d1efafb7234affbd221730310dbaf2f2cfbca272 ? What has changed since then?

sashko: I thought about this and I have no answer. I can only confirm that module is in fact there by default and functions as intended.

martin-ejdestig: Tested on ARP?

PR #380, opened 2019-09-03
neptune3-ui: replace unit file with drop-in

sashko:
Instead of maintaining own unit file, use the original one provided by
neptune3-ui recipe in meta-boot2qt and append/replace PELUX-specific
configuration with systemd drop-in.

Also put required environment variables into a separate file,
/etc/default/neptune.


JEderonn:
Have you tried to override dbus-session and neptune to move them to basic.target and see if it will speed up the boot time?

sashko:
> Have you tried to override dbus-session and neptune to move them to basic.target and see if it will speed up the boot time?

Haven't had a chance yet. It's in my TODO list.

layers/b2qt/recipes-qt/automotive/neptune3-ui_git.bbappend
```diff
@@ -6,8 +6,9 @@
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
-RDEPENDS_${PN}_append = "\
-    dbus-session \
+SRC_URI += "\
+    file://pelux.conf \
+    file://neptune \
```
JEderonn: Cannot see where/how this file is installed into /etc/defaults/

sashko:
Line 31

layers/b2qt/recipes-qt/automotive/neptune3-ui/pelux.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+ExecStart=
+ExecStart=/opt/neptune3/neptune3-ui -r --dbus session -c am-config-neptune.yaml -platform eglfs $EXTRA_ARGUMENTS
+WorkingDirectory=/opt/neptune3
```
JEderonn: Is this needed? Already present in https://code.qt.io/cgit/yocto/meta-boot2qt.git/tree/meta-boot2qt/recipes-qt/automotive/neptune3-ui/neptune.service?h=thud

JEderonn: I also wonder if we should replace
`Environment=AM_STARTUP_TIMER=1`
with
`Environment=AM_STARTUP_TIMER=/tmp/am-startup-timing.log`

PR #382, opened 2019-09-09
Add meta-smarcimx8m-extras

sashko:


martin-ejdestig:
> Snowboy does not provide aarch64 binaries that we can use

Neither of `aarch64-ubuntu1604` nor `android/armv8-aarch64/` can be used? At first I read it as "there are no aarch64 binaries". But it looks like there is. So you tried and it failed?

sashko:
> > Snowboy does not provide aarch64 binaries that we can use
> 
> So you built with one of them and it failed?

Yes, both of the -- linked cannot recognize format in one case, and cannot find some symbols in another.

martin-ejdestig:
OK, perhaps modify the comment in the bbappend a bit then since it states "doesn't provide aarch64 binaries".

meta-smarcimx8m-extras/recipes-alexa/avs-device-sdk/avs-device-sdk_%.bbappend
```diff
@@ -0,0 +1,2 @@
+# build SDK without snowboy support as it doesn't provide aarch64 binaries for Linux
```
martin-ejdestig: This (and the commit message) still reads wrong to me. There are Linux aarch64 binaries (`aarch64` in the name of the folders in snowboy) but you get errors when you try to use them.

martin-ejdestig: Comment here was not changed?

sashko: forgot to git add :man_facepalming:

PR #394, opened 2019-09-18
meson: refresh gtest patch

sashko:
Refreshed the patch added in @e391455d55ad2514fda, so it does not hunk.


sashko:
> Noticed this and let it be. Does it matter in any way?

It is simply annoying as it produces a lot of unnecessary noise in the console.

PR #396, opened 2019-09-20
systemd: use getty-generator for vconsole instantiatiation

sashko:
systemd has a native feature that allows generation and instantiatiation
of serial-getty@.service instances for virtualizer consoles.

### DIFF
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -5,18 +5,10 @@ SRC_URI += "file://99-printk.conf"
 do_install_append() {
     # Make sure resolv.conf is using systemd-resolved
     ln -s /run/systemd/resolve/resolv.conf ${D}${sysconfdir}/resolv.conf
-
-    # Create mutiple consoles
-    tmp="${TTY_LIST}"
-    for ttydev in $tmp ; do
-        ln -sf ${systemd_unitdir}/system/getty@.service \
-            ${D}${sysconfdir}/systemd/system/getty.target.wants/getty@$ttydev.service
-    done
 }
 
 FILES_${PN} += " \
     ${sysconfdir}/resolv.conf \
-    ${sysconfdir}/systemd/system/getty.target.wants \
     ${libdir}/sysctl.d/99-printk.conf \
 "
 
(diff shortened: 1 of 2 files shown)


recipes-core/systemd/systemd_%.bbappend
```diff
@@ -22,6 +15,7 @@ FILES_${PN} += " \
 
 PACKAGECONFIG_remove = " \
     networkd \
+    serial-getty-generator \
```
martin-ejdestig: > systemd has a native feature that allows generation and instantiatiation
> of serial-getty@.service instances for virtualizer consoles.

Hum, I do not understand. This line makes it so that it is not used?

martin-ejdestig: Or this is two changes in one? One that remove our own `ln -s` for getty@.service and relies on `systemd-getty-generator`. And the PACKAGECONFIG_removes generation of serial gettys... ?

martin-ejdestig: `serial-getty-generator` PACKAGECONFIG is disabled by default? From systemd recipe:

```
# Use the upstream systemd serial-getty@.service and rely on
# systemd-getty-generator instead of using the OE-core specific
# systemd-serialgetty.bb - not enabled by default.
PACKAGECONFIG[serial-getty-generator] = ""
```

martin-ejdestig:
@0390b8ae1061dceaa8cb97f13870fd9979c009a6 also added `getty.target.wants` to `FILES_${PN}`.

martin-ejdestig:
I wonder why `serial-getty-generator` did not dynamically create getty services before in PELUX. It has been a feature in systemd for a very long time (the start of the project almost... to create them on demand and saving boot time in that way.

But if it works now, let's just remove this and be happy that we remove some services from the startup. :)

sashko:
@SKumarDas do you remember why this change was brought in?

SKumarDas:
@sashko This change was made to facilitate one more console because before this change we had only one and it was taken by the GUI.

sashko:
> No time to test this myself, but approving assuming we still get virtual consoles.

Feel free to come by my desk and check it out on NUC or Raspberry Pi I have here.

PR #400, opened 2019-09-26
psplash: add systemd unit files

sashko:
Added unit files for two services:
- psplash-start that will start psplash on system boot
- psplash-stop that will stop it right after Neptune 3 UI is started,
  when user switches virtual consoles, user prompt is shown instead of
  the splash screen.

At the moment, default OpenEmbedded screen is shown.

### DIFF
--- a/recipes-core/psplash/psplash_%.bbappend
+++ b/recipes-core/psplash/psplash_%.bbappend
@@ -1,0 +1,17 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += " \
+    file://psplash-start.service \
+    file://psplash-stop.service \
+"
+
+inherit systemd
+
+SYSTEMD_PACKAGES = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', '${PN}', '', d)}"
+SYSTEMD_SERVICE_${PN}_append = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', ' psplash-start.service psplash-stop.service', '', d)}"
+
+do_install_append () {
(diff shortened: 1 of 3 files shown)


recipes-core/psplash/psplash_%.bbappend
```diff
@@ -0,0 +1,17 @@
...
+"
+
+inherit systemd
+
+SYSTEMD_PACKAGES = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', '${PN}', '', d)}"
+SYSTEMD_SERVICE_${PN}_append = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', ' psplash-start.service', '', d)}"
+
+do_install_append () {
+    install -d ${D}${systemd_unitdir}/system
+    install -m 644 ${WORKDIR}/psplash-start.service ${D}${systemd_unitdir}/system
```
JEderonn: consider to use `${systemd_system_unitdir}` instead of `${systemd_unitdir}/system`

recipes-core/psplash/psplash_%.bbappend
```diff
@@ -0,0 +1,17 @@
...
+
+SRC_URI += " \
+    file://psplash-start.service \
+    file://psplash-stop.service \
+"
+
+inherit systemd
+
+SYSTEMD_PACKAGES = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', '${PN}', '', d)}"
+SYSTEMD_SERVICE_${PN}_append = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', ' psplash-start.service', '', d)}"
```
JEderonn: shall psplash-stop.service also be appended?

PR #401, opened 2019-09-26
swupdate: disable systemd service

sashko:
swupdate is currently configured to use one of the local development
workstation as a server which is not always available.

Moreover, we do not want swupdate to check for any updates on start, as
it is more of a proof of concept than a production ready solution, so
disable the corresponding systemd service by default.

It was once disabled in 73c3163ef5, but was later reenabled in
6f7e4f29e6.


martin-ejdestig:
> It was once disabled in @73c3163ef5, but was later reenabled in @6f7e4f29e6.

Just an FYI: If you do not add the @ commit hashes will be turned into clickable links on github. (gitk makes the hashes clickable regardless, but nice to be able to click links directly in the browser as well.)

PR #405, opened 2019-10-01
layer.conf: use BBFILES_DYNAMIC to handle dynamic-layers

sashko:
BBFILES_DYNAMIC is a native bitkbake feature that can be used to
activate content only when some other layers are present. The other
layers are identified by the collections that they define.

Use this feature instead of manually going through all the recipes
and appends and adding them only if the corresponding layer is
included in our configuration.

Also renamed layers to dynamic-layers as it is a common practice
to assign this name to the folder in the Yocto community.

For more info see 337bea6573 in poky or the Yocto Project Reference
Manual[1].

[1] https://yoctoproject.org/docs/latest/ref-manual/ref-manual.html#var-BBFILES_DYNAMIC


conf/layer.conf
```diff
@@ -15,13 +15,12 @@ BBFILE_PATTERN_pelux-layer := "^${LAYERDIR}/"
...
 
-# Let us add layer-specific bbappends which are only applied when that
-# layer is included in our configuration
-BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bbappend' % layer \
-               for layer in BBFILE_COLLECTIONS.split())}"
-# Do the same for recipes
-BBFILES += "${@' '.join('${LAYERDIR}/layers/%s/recipes*/*/*.bb' % layer \
-               for layer in BBFILE_COLLECTIONS.split())}"
+BBFILES_DYNAMIC += " \
+    b2qt:${LAYERDIR}/dynamic-layers/b2qt-layer/*/*/*.bb \
```
JEderonn: consider rename `b2qt-layer` to `meta-boot2qt` to be consistent with the naming of the parent layer.

sashko: this seems to be an established syntax:
- https://github.com/agherzan/meta-raspberrypi/tree/master/dynamic-layers
- https://github.com/Freescale/meta-freescale/tree/master/dynamic-layers

JEderonn: I had a look into the reference manual: https://yoctoproject.org/docs/latest/ref-manual/ref-manual.html#var-BBFILES_DYNAMIC
and you can see that the parent layer for `clang-layer` is `meta-clang`, and for `core` it's `openembedded-core`, and not `clang-layer` and `core-layer`, that's why I commented this.

But I'm fine with `xx-layer` instead of `meta-xx`.

However, I think `boot2qt-layer` is more appropriate than `b2qt-layer? , don't you think?

sashko: I believe it comes from `BBFILE_COLLECTIONS` of the layer.

PR #417, opened 2019-10-04
variant/arp-intel: remove unnecessary machine conf inclusion

sashko:


JEderonn:
How and where is the machine configuration included if not from these files?
Perhaps it's worth to mention in the commit message.

PR #426, opened 2019-11-18
qtapplicationmanager-sc.inc: Replace += with _append

JEderonn:
If using += the behaviour is not defined in case the variable
hasn't been set before bitbake parses this line. There is a risk
that the variable is (re)set in the main recipe if _append is
not used.

Signed-off-by: Johan Ederonn <jederonn@luxoft.com>

PR #427, opened 2019-11-22
WIP: Restructure variant

DunderRoffe:
This is a suggestion on how to restructure the variant system to reduce duplication when adding new variants and hardware. I do __NOT__ want this to be merged. I just want feedback.

### Overview
In this setup we only have special files containing bblayers for bsp or features (such as QtAuto). 
Everything else is either located in sample/local.conf.sample or classes/core-image-pelux.bbclass

### Benefits of this solution
Meaning using this setup it is possible to add support for a new hardware and only have to add one bblayers-new-hw.conf file in order for the new hardware to be able to use all features (such as qt-auto). The same goes for features (since they are basically the same thing).

Easier to understand what is tweakable as all is in the same local.conf file

### Drawbacks of this solution are:
Two additional manual steps are now required to setup pelux bitbake:
* Uncommenting the MACHINE to use in conf/local.conf
* Uncommenting bsp and features in conf/bblayers.conf

### Unknowns
This solution does not offer the possibility to treat arp as a separate thing out of the box as there is no MACHINE called arp (and there should really not be either). One idea to solve this is to treat arp as a "feature" and added using dynamic-layers when meta-arp is present like we do with QtAuto.

### Required future work:
* Rewrite CI scripts and rethink some of the design to be able to do the extra manual steps mentioned above. Some of this is already done though with ACCEPT_EULA
* Rewrite documentation for how to build.


JEderonn:
In general, good and easy to understand.

Consider to either move the bblayers-<MACHINE>.conf to a separate sub-folder or rename these files to bblayers-machine-<MACHINE>.conf or bblayers-bsp-<MACHINE>.conf to indicate which category the file belongs to.

conf/variant/common/conf-notes.txt
```diff
@@ -0,0 +1,12 @@
...
+1. Set MACHINE in "conf/local.conf"
+2. Enable BSP and or other features such as QtAuto in "conf/bblayers.conf"
+
+Available PELUX images:
+* core-image-pelux-minimal
+* core-image-pelux-minimal-update
+
+* core-image-pelux-qtauto-neptune     (Requires QtAuto)
+
+Build them by running: bitbake <image>
```
martin-ejdestig: This will be printed after source:ing... but it is OK to modify conf/bblayers.conf after that and bitbake will do the correct thing? One can't change things in `conf/` that would require "re-source:ing"? I assume it is OK, just checking.

DunderRoffe: sourcing just copies from `source/meta-pelux/conf/.../*.conf.sample` to `build/conf/*conf`
Meaning editing `build/conf/*conf` after sourcing is fine. No resourcing required

Even better this is the intended use and resourcing will __not__ have any effect (besides setting the build env and cd'ing to the build directory). To reset one needs to `rm build/conf/templateconf.cfg`

martin-ejdestig:
(This should be a comment on the commit message. Github can not handle that, it seems.)

> Restructure conf/variant/common -> conf/sample 

Is it not a bit weird to call all our configuration's "samples"? I do not have a better suggestion though and we called them samples before as well. Or is it something from the bitbake world? Hum... feel free to ignore this comment. :)

DunderRoffe:
It's from the bitbake world. The files have to be called bitbake.conf.sample and local.conf.sample

classes/core-image-pelux.bbclass
```diff
@@ -6,7 +6,7 @@
 
 inherit core-image
 
-IMAGE_FEATURES += "package-management debug-tweaks"
+IMAGE_FEATURES_append = " package-management debug-tweaks"
```
martin-ejdestig: Is this due to something similar to https://github.com/Pelagicore/meta-pelux/pull/426 ?

conf/variant/common/conf-notes.txt
martin-ejdestig: OK, good.

Another thing just occurred to me though. How do we do this in a simple way in the CI?

`$ sed -i 's/# MACHINE=foo/MACHINE=foo/` conf/local.conf`

martin-ejdestig: Can the correct thing be done in `conf/bblayers.conf` depending on `MACHINE` perhaps? I do not know in which order things are parsed.

classes/core-image-pelux.bbclass
DunderRoffe: Yes.

Although I do not remember if it was a problem, or if it was for future-proofing / making it look like other places where we have changed to use _append instead of += ...

Most likely the latter.

martin-ejdestig: OK, maybe just mention "+= -> _append for IMAGE_FEATURES to be consistent" in commit message.

conf/variant/common/conf-notes.txt
DunderRoffe: > Can the correct thing be done in `conf/bblayers.conf` depending on `MACHINE` perhaps? I do not know in which order things are parsed.

No. AFAIK bblayers.conf does not share the same context as local.conf

DunderRoffe: > OK, good.
> 
> Another thing just occurred to me though. How do we do this in a simple way in the CI?
> 
> ```shell
> $ sed -i 's/# MACHINE=foo/MACHINE=foo/` conf/local.conf
> $ sed -i 's/# require conf/sample/bar.conf/require conf/sample/bar.conf` conf/bblayers.conf
> ```
> 
> Ugh. :/

Uncommenting is already done by CI for ACCEPT_EULA when building for jetson-tx2

DunderRoffe: And CI knows what machine is supposed to be used.

But it is most likely wise to think a bit about this before implementing

DunderRoffe:
rebased on master

conf/sample/bblayers-jetson-tx2.conf
```diff
@@ -0,0 +1,4 @@
+BBLAYERS += "\
+    ${BSPDIR}/sources/meta-tegra \
+    ${BSPDIR}/sources/meta-boot2qt/meta-tegra-extras \
```
JEderonn: When meta-boot2qt is bumped to QtAS_5.14.0 (warrior) iof 5.13.1 (thud) the meta-xxx-extras BSP layers have been replaced with dynamic-layers in meta-boot2qt-distro.
But I assume this needs to be kept if meta-pelux shall be compatible with thud and previous Yocto releases.

DunderRoffe: Yes you are correct.
I think it will be tricky to support both warrior and thud at the same time. At least for qt.
Also do we really wanna keep using meta-boot2qt-distro?
What parts of meta-tegra-extras are actually used?

JEderonn: I would really like to see that we get rid of the dependency to meta-boot2qt-distro.
I have not investigated which part of the meta-xxx-extras that are used.
But will there be a problem if we keep the meta-xxx-extras layer in BBLAYER even if they don't exist? When we support zeus we can remove it, but we should strive to support at least 2 Yocto versions, w and w/o Qt.

conf/variant/common/conf-notes.txt
JEderonn: Can you create a new Jenkins job which evaluates the new way of configuring and see how it works?

conf/sample/bblayers-jetson-tx2.conf
DunderRoffe: AFAIK bitbake will raise sanity check errors when parts of BBLAYER does not exist. Will do a test...

conf/variant/common/conf-notes.txt
DunderRoffe: I will make an attempt. Creating a WIP PR for pelux-manifests

DunderRoffe:
See https://github.com/Pelagicore/pelux-manifests/pull/474 for suggestion on how to do CI for this solution

DunderRoffe:
Closing due to inactivity

PR #430, opened 2019-11-27
/etc/defaults/neptune: Change XDG_RUNTIME_DIR

JEderonn:
There is a change in Qt 5.14, where the XDG_RUNTIME_DIR cannot be
/tmp since the permission for that dir is not read-only for the
current user (i.e. root). The default value of XDG_RUNTIME_DIR as
defined in meta-boot2qt is /run/user/0 so let's use this also.

TODO: Consider to reuse the default Qt config as defined in the
boot2qt addons and only apply modifications applicable for PELUX.

Signed-off-by: Johan Ederonn <jederonn@luxoft.com>

### DIFF
--- a/dynamic-layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune
+++ b/dynamic-layers/b2qt/recipes-qt/automotive/neptune3-ui/neptune
@@ -2,7 +2,7 @@ HOME=/home/%u/
 LC_ALL=en_US
 DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/0/dbus_session_socket
 QT_IM_MODULE=qtvirtualkeyboard
-XDG_RUNTIME_DIR=/tmp
+XDG_RUNTIME_DIR=/run/user/0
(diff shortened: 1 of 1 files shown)


JEderonn:
> Some more information (e.g. error message, where in Qt the problem is or e.g. a reference to a Git commit in Qt/meta-boot2qt that might have more info) would be nice to be able to verify. The commit message now seems a bit strange to me without this.

I have added the generated log and the hash in qtbase where the change was made.

> E.g. I do not think it makes sense for XDG_RUNTIME_DIR to be read-only. Is this a typo? Looking up what the spec, https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html , says I think it is. Or what is going on?

The commit message states "Conform to XDG Base Directory Specification"

> Btw, when is `/run/user/0` created now and by whom? (Normally done by `logind` on a "normal" Linux system, I think, ... or some other login manager if systemd is not used... but it is not run in PELUX, nor do we create a session for Neptune.) I assume you have tested this after a clean boot and not created the directory manually and just restarted Neptune?

I have not analyzed who creates the dir. But yes, I have tested this on clean boot. This is also how the boot2qt distro is setup.

PR #431, opened 2019-12-02
pelux.conf: Temporarily disable DISTRO_FEATURE process-containment

JEderonn:
Due to some issues which cannot be fixed right now:
- (BLOCKER) Qt 5.14 AS candidate results in build error
  ERROR: Required build target 'neptune3-ui' has no buildable providers.
  Missing or unbuildable dependency chain was: [
    'neptune3-ui',
    'qtapplicationmanager-native',
    'qtapplicationmanager-native-softwarecontainer-native']
- With Qt 5.13, Neptune3-UI cannot run applications in container
  (workaround is to not provide sc-config.yaml to neptune3-ui)
- Sometimes at system shutdown, the softwarecontainer-agent does
  not stop within maximum timeout

### DIFF
--- a/conf/distro/pelux.conf
+++ b/conf/distro/pelux.conf
@@ -28,6 +28,22 @@ DISTRO_FEATURES_append = " \
     webengine\
 "
 
+# Temporarily disable process-containment due to some issues which
+# cannot be fixed right now:
+#  - (BLOCKER) Qt 5.14 AS candidate results in build error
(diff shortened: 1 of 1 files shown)


JEderonn:
> I am perhaps a bit worried that "why" is not obvious to someone looking in the file. But could always `git blame` I  guess.

You mean that you would like to improve the comments in the file? I can do that.

PR #432, opened 2019-12-02
qtivi-mopidy-plugin_git.bb: Add suffix _simulation to lib

JEderonn:
The intention has been that the QtIvi Media mopidy plugin shall
not be selected since it's still experimental and not complete.
The problem is that it is still selected due to incorrect naming.

The method QIviServiceManagerPrivate::isSimulation returns TRUE
if the QtIvi plugin library name contains _simulation.


JEderonn:
Actually, I agree with you that it would be better to not have it installed at all, and at the same time rename it back to it's original naming since it's not really a simulation backend.

PR #436, opened 2020-01-09
Update PELUX psplash logo

DunderRoffe:
Change psplash logo to a one that looks good in both portrait and
landscape mode.


recipes-core/psplash/psplash/psplash-start.service
```diff
@@ -5,7 +5,7 @@ After=systemd-vconsole-setup.service systemd-udev-trigger.service systemd-udevd.
 DefaultDependencies=no
 
 [Service]
-ExecStart=/usr/bin/psplash --angle 270
+ExecStart=/usr/bin/psplash
```
martin-ejdestig: OK to remove because you rotated the image itself?

DunderRoffe: The image is symetrical. It looks identical from all sides.

DunderRoffe: Still testing though ... plz wait with merging

martin-ejdestig: Commit message updated with this info so resolving.

DunderRoffe:
Tested and seems to work nice

DunderRoffe:
> Tested and seems to work nice

Nvm could do some more tweaking

JEderonn:
If you consider tweaking is finished, pl merge.

DunderRoffe:
Tweaking done

PR #296, opened 2019-04-23
Acrn

DunderRoffe:
This PR adds config and recipes used to create Service-OS and User-OS images using PELUX.

It should be fairly non-intrusive and is tested and verified to work with ARP/Intel. However, there are some things to consider:

* It is based on poky warrior since some things in the setup is broken in thud
* There is a workaround commit (104f668) that is not compliant with swupdate so this needs to be fixed

As this is already tested I am mostly interested in semantics, the notes above and or structure reviews.

PR #437, opened 2020-01-14
Add build configuration Acrn

DunderRoffe:
This PR adds config and recipes used to create Service-OS and User-OS images using PELUX.

It should be fairly non-intrusive and is tested and verified to work with ARP/Intel. However, there are some things to consider:

    It is based on poky warrior since some things in the setup is broken in thud

This PR is a rebase of #296 on latest master as of 2020-01-14
Workarounds used in that PR are now obsolete so should be fine to merge


DunderRoffe:
None of the changes above break anything existing.
These are recipes and config to be used with meta-acrn

PR #438, opened 2020-01-23
core-image-pelux.bbclass: Add mosquitto

JEderonn:


classes/core-image-pelux.bbclass
```diff
@@ -48,6 +48,7 @@ IMAGE_INSTALL_append = "\
     mopidy \
     systemd-resource-control \
     user-identification-manager \
+    mosquitto \
```
martin-ejdestig: Nit: sort?

JEderonn: fixed

